PR #5, opened 2024-08-27
update the code so it can be compiled with latest dependencies

ngeojiajun:


Sn1F3rt (approved):

R0BC0D3R:
Confirmed that the code compiles and runs on Windows and Linux when using latest dependencies.

PR #20, opened 2024-10-14
build: fix mac arm build

Sn1F3rt:
- rebase CMakeLists.txt from Monero upstream 
- update documentation


old-greywolf:
good job, Sn1F3rt.

PR #31, opened 2024-10-16
build(depends): drop aes support for android arm

Sn1F3rt:


Sn1F3rt:
Does not fix. To be looked into at a later date.

PR #32, opened 2024-11-22
chore(mms): don't show error when no mms file is available

Sn1F3rt:


src/wallet/message_store.cpp
```diff
@@ -725,7 +725,7 @@ void message_store::read_from_file(const multisig_wallet_state &state, const std
   {
     // Simply do nothing if the file is not there; allows e.g. easy recovery
     // from problems with the MMS by deleting the file
-    MERROR("No message store file found: " << filename);
+    MINFO("No message store file found: " << filename);
```
ngeojiajun, line 728: Any particular reason for this change?

ngeojiajun: Please disregard this, it looks like a fix from upstream

ngeojiajun (approved):

PR #70, opened 2026-04-06
fix: bound wallet scan threadpool to half of available cores

sn1f3rt:
## Summary

- Replaces the dedicated `m_scan_threadpool` with a counting semaphore (`m_scan_semaphore`) over the global `threadpool::getInstance()` — eliminates the N×1.5 total thread count that arose from having two separate pools
- Semaphore is initialised to `max(1, get_max_concurrency() / 2)` at wallet construction, after `wallet_args` has called `set_max_concurrency()` from `--max-concurrency`, so the flag is correctly respected
- CPU-intensive leaf tasks (output checking, block/tx parsing, cache and key derivation work) go through `scan_submit()`, which wraps each with a RAII scope guard around semaphore acquire/release — the guard ensures the slot is always returned even if the task throws
- `pull_and_parse_next_blocks` is submitted without the semaphore since it is an I/O-bound orchestrator that internally dispatches those leaf tasks; wrapping it would cause a slot-hold deadlock on single-core machines where tasks execute inline
- Caps each `getblocks.bin` response to `BLOCKS_SYNCHRONIZING_DEFAULT_COUNT` (100) blocks via the `max_count` request field, down from the server default of 1000 — reduces `m_daemon_rpc_mutex` hold time per batch
- Fixes the macOS `DEFAULT_MAX_CONCURRENCY=1` pthread workaround in `wallet_args.cpp` — the previous `is_arg_defaulted` guard meant the default was never applied unless the user explicitly passed `--max-concurrency`

## Relationship to #77

#77 (now merged) fixed RPC unresponsiveness by moving `refresh()` to a dedicated background thread. This PR is independent and addresses the remaining issues #77 did not cover: CPU saturation from uncapped scan parallelism, the 1000-block batch size on the wallet RPC endpoint, and the macOS concurrency default bug.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2476,6 +2476,7 @@ void wallet2::pull_blocks(uint64_t start_height, uint64_t &blocks_start_height, 
   req.prune = true;
   req.start_height = start_height;
   req.no_miner_tx = m_refresh_type == RefreshNoCoinbase;
+  req.max_count = BLOCKS_SYNCHRONIZING_DEFAULT_COUNT;
   {
     const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
     bool r = net_utils::invoke_http_bin("/getblocks.bin", req, res, m_http_client, rpc_timeout);
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1144,6 +1144,8 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended):
   m_rpc_version(0),
   m_export_format(ExportFormat::Binary)
 {
+  unsigned scan_threads = std::max(1u, boost::thread::hardware_concurrency() / 2);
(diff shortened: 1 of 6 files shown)


src/wallet/wallet2.cpp
```diff
@@ -1142,14 +1142,24 @@ wallet2::wallet2(network_type nettype, uint64_t kdf_rounds, bool unattended):
...
 wallet2::~wallet2()
 {
 }
 
+void wallet2::scan_submit(tools::threadpool &tpool, tools::threadpool::waiter *waiter, std::function<void()> f, bool leaf)
+{
+  tpool.submit(waiter, [this, f = std::move(f)]() {
+    m_scan_semaphore.acquire();
+    f();
+    m_scan_semaphore.release();
```
ngeojiajun: any guarantee that f() will not throw?

sn1f3rt: Good catch — fixed in 9e8bb7b. Replaced the bare `release()` with a `create_scope_leave_handler` RAII guard so the slot is released regardless of whether `f()` throws.

ngeojiajun (approved):
this seems fine, but a proper testing is required

PR #77, opened 2026-04-23
perf: wallet rpc responsiveness

### LINKED ISSUE #66: investigate: inconsistent wallet scan performance
## Description

Community reports of inconsistent wallet scan behaviour across different builds. Two distinct modes have been observed:

1. High CPU usage — RPC becomes unresponsive, scan completes in several minutes
2. Low CPU usage — RPC remains responsive, but scan takes over an hour

The root cause is unknown. It is unclear what build conditions or configuration determine which behaviour is exhibited.

## Steps to Investigate

- [ ] Identify what differs between builds that produce each behaviour
- [ ] Profile CPU usage and RPC responsiveness during wallet scan
- [ ] Check for any threading or priority changes affecting scan vs RPC concurrency
- [ ] Determine if there is a way to improve scan speed while keeping RPC responsive

## References

Community report from R0BC0D3R on Discord.

R0BC0D3R:
Keep wallet-rpc responsive during restore/sync
refresh() previously ran on the HTTP thread, blocking all RPC calls
for the duration of a wallet restore. Port simplewallet's idle thread
pattern: refresh runs in a dedicated background thread; handlers call
stop() to yield the current batch before locking, keeping read RPCs
responsive throughout restore.

Closes #66


ngeojiajun (approved):

src/wallet/wallet2.cpp
```diff
@@ -2562,17 +2562,28 @@ void wallet2::process_parsed_blocks(uint64_t start_height, const std::vector<cry
...
-        gender(iod);
+      if (is_hw_device)
+      {
+        boost::unique_lock<hw::device> hwdev_lock(hwdev);
+        for (auto &iod: slot.primary)
+          gender(iod);
+        for (auto &iod: slot.additional)
+          gender(iod);
+      }
+      else
```
sn1f3rt: `else` branch is identical to the `if` branch minus the lock. The two loops could live outside the conditional: `if (is_hw_device) { lock; } for primary... for additional...` — avoids duplicating the iteration.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -97,13 +97,34 @@ namespace
...
+#define LOCK_IDLE_SCOPE() \
+  if (m_wallet) m_wallet->stop(); \
+  boost::unique_lock<boost::mutex> idle_guard_(m_idle_mutex); \
+  epee::misc_utils::auto_scope_leave_caller idle_notify_ = \
+    epee::misc_utils::create_scope_leave_handler([&]{ m_do_refresh.store(true, std::memory_order_relaxed); m_idle_cond.notify_all(); })
+
+// LOCK_IDLE_SCOPE_RO: stops background refresh (same as LOCK_IDLE_SCOPE) and restarts it on exit.
+// The idle thread holds the mutex for the entire refresh() call, so there is no "gap" to wait for.
+// Calling m_wallet->stop() causes refresh to exit after the current batch (~5-500ms), keeping
+// RPC latency low. Refresh restarts immediately on scope exit via m_do_refresh=true.
```
sn1f3rt: `LOCK_IDLE_SCOPE_RO` and `LOCK_IDLE_SCOPE` expand to identical code — both call `m_wallet->stop()` and restart refresh on scope exit. The names imply a distinction (read-only vs write) that does not exist in the implementation. Either differentiate them (e.g. RO does not call `stop()` and just waits for the mutex) or collapse into one macro to avoid misleading future readers.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -2948,15 +3061,27 @@ namespace tools
...
+        m_idle_cond.notify_all();
+      }
+      // Wait for completion. condition_variable::wait releases the mutex while
+      // sleeping, so IO thread 2 can serve get_balance/get_height between batches.
+      boost::unique_lock<boost::mutex> lock(m_idle_mutex);
+      m_refresh_done_cond.wait(lock, [this]{
+        return !m_do_refresh.load(std::memory_order_relaxed) ||
+               !m_idle_run.load(std::memory_order_relaxed);
+      });
       return true;
```
sn1f3rt, line 3076: This always returns `true`, even if the idle thread's refresh threw an exception. The idle thread logs the error but does not propagate it, so the RPC caller gets a success response on failure. The original code caught the exception and returned `false` with an error. Consider storing the exception (e.g. via `std::exception_ptr`) in a shared field and re-throwing or returning an error here.

src/wallet/wallet2.cpp
R0BC0D3R: Fixed using boost::defer_lock — lock only acquired when is_hw_device is true, loops moved outside the conditional.

src/wallet/wallet_rpc_server.cpp
R0BC0D3R: Collapsed into one macro. We previously tried differentiating them but skipping stop() for read handlers reintroduces the latency problem — the idle thread holds the mutex for the full refresh() call so stop() is required regardless. Macro comment updated to explain this.

src/wallet/wallet_rpc_server.cpp
R0BC0D3R: Fixed via std::exception_ptr m_refresh_exception (protected by m_idle_mutex). Idle thread clears it before each refresh and sets it on exception. on_refresh checks it after the wait and returns false with an error on failure.

sn1f3rt (approved):
All three issues addressed. LGTM.

PR #81, opened 2026-05-15
fix: p2p security and privacy improvements

R0BC0D3R:
## Summary

Ports several security and privacy fixes from upstream Monero to `src/p2p/net_node.inl` and `src/p2p/net_node.h`.

- **Anonymous inbound address leak** (Monero PR #9632): Self-address inserted at a random position with timestamp=0 in timed sync responses, preventing correlation of our own onion/I2P address by list position or wall-clock time across responses.

- **Peer diversity** (Monero PR #9939): Upgraded subnet deduplication from /16 to /24, and changed the tried-peers tracking set from list index to peer ID to avoid skipping valid peers when the peerlist is modified mid-iteration.

- **Stale outgoing connection counter**: `get_outgoing_connections_count` now refreshes `m_current_number_of_out_peers` on every call, preventing the connection-maker from stalling when peers drop between 1-second poll intervals.

- **Peer ID disabled over anonymity networks** (Monero commit 37bb594): Peer IDs were randomly assigned to all network zones including Tor and I2P, allowing an observer to correlate connections across anonymity network traffic. Fixed by defaulting peer_id to 1 for non-public zones and guarding all peer_id comparisons with a public-zone check. Anonymity networks now identify duplicate connections by address only.

- **Silent tx drop on privacy networks** (Monero PR #9267): Tx send no longer selects an anonymity network zone that has no active outgoing connections, preventing transactions from being silently queued with no chance of delivery.

## Test plan

- [ ] Builds cleanly on target platforms
- [ ] Daemon starts and connects to peers normally
- [ ] No regression in outbound connection count over time

### DIFF
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -2329,7 +2329,14 @@ namespace nodetool
     real peer with that identity banned/blacklisted. */
 
     if(!context.m_is_income && zone.m_our_address.get_zone() == zone_type)
-      rsp.local_peerlist_new.push_back(peerlist_entry{zone.m_our_address, zone.m_config.m_peer_id, std::time(nullptr)});
+    {
+      // Random position + timestamp=0: prevents correlating our own onion/I2P address
+      // by its list position or a real wall-clock timestamp across responses (Monero PR #9632).
+      rsp.local_peerlist_new.insert(
+        rsp.local_peerlist_new.begin() + crypto::rand_range(std::size_t(0), rsp.local_peerlist_new.size()),
+        peerlist_entry{zone.m_our_address, zone.m_config.m_peer_id, 0}
+      );
+    }
 
     LOG_DEBUG_CC(context, "COMMAND_TIMED_SYNC");
     return 1;
--- a/src/p2p/net_node.inl
+++ b/src/p2p/net_node.inl
@@ -1395,6 +1395,10 @@ namespace nodetool
         continue;
       }
 
+      if(m_network_zones.at(pe.adr.get_zone()).m_our_address == pe.adr) {
+        continue;
+      }
+
       if(!is_remote_host_allowed(pe.adr)) {
         continue;
       }
@@ -1433,8 +1437,8 @@ namespace nodetool
       const uint32_t next_needed_pruning_stripe = m_payload_handler.get_next_needed_pruning_stripe().second;
 
       // build a set of all the /24 subnets we're connected to, and prefer a peer that's not in that set
-      // subnet_mask is the /24 mask in host byte order (0x00ffffff on little-endian, Monero PR #9939)
-      const uint32_t subnet_mask = 0x00ffffff;
+      // m_ip is stored in network byte order; 0xffffff00 preserves the first three octets (/24)
+      const uint32_t subnet_mask = 0xffffff00;
       std::set<uint32_t> connected_subnets;
       if (&zone == &m_network_zones.at(epee::net_utils::zone::public_)) // at returns reference, not copy
       {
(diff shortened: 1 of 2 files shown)


sn1f3rt (commented):
~ignore~

sn1f3rt (commented):
Test comment.

sn1f3rt (commented):
Test comment.

sn1f3rt (commented):
Test

src/p2p/net_node.inl
```diff
@@ -1427,18 +1432,19 @@ namespace nodetool
       size_t random_index;
       const uint32_t next_needed_pruning_stripe = m_payload_handler.get_next_needed_pruning_stripe().second;
 
-      // build a set of all the /16 we're connected to, and prefer a peer that's not in that set
-      std::set<uint32_t> classB;
+      // build a set of all the /24 subnets we're connected to, and prefer a peer that's not in that set
+      // subnet_mask is the /24 mask in host byte order (0x00ffffff on little-endian, Monero PR #9939)
+      const uint32_t subnet_mask = 0x00ffffff;
```
sn1f3rt: The mask `0x00ffffff` is incorrect for `m_ip`, which is stored in network byte order (as returned by `inet_addr()`). In network byte order, for an IP like `1.2.3.4`, the `uint32_t` value is `0x01020304` — so a /24 mask preserving the first three octets needs `0xffffff00`, not `0x00ffffff`. With `0x00ffffff`, you're masking the last three octets (the host part), so two addresses in the same /24 produce different masked values and won't be deduplicated. The same byte-order bug existed in the pre-existing /16 mask (`0x0000ffff` should be `0xffff0000`). Both should be corrected here.

sn1f3rt (commented):
Two inline issues identified in the diff.

src/p2p/net_node.inl
```diff
@@ -1314,6 +1316,9 @@ namespace nodetool
     if (zone.m_connect == nullptr)
       return false;
 
+    if (zone.m_our_address == na)
```
sn1f3rt, line 1319: The self-address guard added to `make_new_connection_from_peerlist` (checking `zone.m_our_address == pe.adr`) is missing in `make_new_connection_from_anchor_peerlist`. If our own address ever appears in the anchor peerlist (e.g. after a restart with a stale anchor list), that path will attempt a self-connection. The guard in `check_connection_and_handshake_with_peer` provides a backstop, but defence-in-depth there would be consistent with the fix applied to the other peerlist path.

src/p2p/net_node.inl
R0BC0D3R: Fixed in 8a68658: Corrected to 0xffffff00. m_ip is in network byte order so 0x00ffffff was masking the last three octets rather than the first, making /24 deduplication non-functional. Updated the comment accordingly.

src/p2p/net_node.inl
R0BC0D3R: Fixed in 8a68658: Added m_network_zones.at(pe.adr.get_zone()).m_our_address == pe.adr check to make_new_connection_from_anchor_peerlist, matching the pattern already present in make_new_connection_from_peerlist.

sn1f3rt (approved):
LGTM. The critical subnet mask byte-order bug ( → ) has been correctly fixed. The remaining two observations (write-back in  and unguarded seed/priority peer call sites) are not blockers — the counter is atomic so there's no data race, and the unguarded paths require operator misconfiguration to trigger. Good to merge.

sn1f3rt (approved):
LGTM. The critical subnet mask byte-order bug has been correctly fixed. The remaining two observations (write-back in get_outgoing_connections_count and unguarded seed/priority peer call sites) are not blockers — the counter is atomic so there is no data race, and the unguarded paths require operator misconfiguration to trigger. Good to merge.

ngeojiajun (approved):

PR #82, opened 2026-05-15
fix: P2P message deserialization and packet size hardening

R0BC0D3R:
## Summary

Backports upstream Monero security mitigations for P2P denial-of-service
attack vectors. Changes are split into two areas:

**Portable storage deserialization limits** (epee)
- Adds a global object field counter (`m_fields`) separate from the object
  count, capped at 262,144 fields per message
- Rejects duplicate keys within a section
- Rejects unnamed section entries (zero-length key)
- Bounds-checks section arrays against the per-message object limit

**Levin protocol packet size controls**
- New connections are limited to 256 KiB until the P2P handshake completes,
  then upgraded to the full 100 MB cap
- Per-command byte limits are set on each new connection (e.g. 4 KiB for
  PING/support-flags, 64 KiB for handshake/timed-sync, 128 MiB for
  block and transaction responses)
- Transaction flood notifications are skipped for peers that have not yet
  completed the handshake

## Test plan
- [ ] Daemon starts and forms peer connections without packet-size warnings
- [ ] Blockchain sync proceeds normally
- [ ] Transaction send/receive works correctly
- [ ] No regressions in peer connectivity or mining


ngeojiajun (approved):

contrib/epee/include/net/levin_base.h
```diff
@@ -72,6 +72,7 @@ namespace levin
 
 
 #define LEVIN_DEFAULT_TIMEOUT_PRECONFIGURED 0
+#define LEVIN_INITIAL_MAX_PACKET_SIZE 256*1024      //256 KiB before handshake
```
sn1f3rt: Macro arithmetic should be parenthesized:
```c
#define LEVIN_INITIAL_MAX_PACKET_SIZE (256*1024)
```
Without parentheses, use in a compound expression (e.g. `X / LEVIN_INITIAL_MAX_PACKET_SIZE`) would evaluate as `X / 256 * 1024` due to left-to-right associativity. Harmless in the current usages but a latent footgun.

contrib/epee/include/net/levin_protocol_handler_async.h
```diff
@@ -577,10 +582,13 @@ class async_protocol_handler
...
           m_oponent_protocol_ver = m_current_head.m_protocol_version;
-          if(m_current_head.m_cb > m_config.m_max_packet_size)
+          if (m_connection_context.handshake_complete())
+            m_max_packet_size = m_config.m_max_packet_size;
+          const size_t max_bytes = m_connection_context.get_max_bytes(m_current_head.m_command);
+          if(m_current_head.m_cb > std::min<uint64_t>(m_max_packet_size.load(std::memory_order_relaxed), max_bytes))
           {
-            LOG_ERROR_CC(m_connection_context, "Maximum packet size exceed!, m_max_packet_size = " << m_config.m_max_packet_size 
-              << ", packet header received " << m_current_head.m_cb 
+            LOG_ERROR_CC(m_connection_context, "Maximum packet size exceed!, m_max_packet_size = " << max_packet_size
```
sn1f3rt: `max_packet_size` here is the local captured at the top of `handle_recv` (before the upgrade on line 585). On the first packet received after handshake completion, the atomic is upgraded to `m_config.m_max_packet_size` (100 MB) and the size check on line 588 uses that fresh value — but the log still prints the old pre-handshake limit (256 KiB). The message will mislead anyone diagnosing a drop on that first post-handshake packet.

Fix: reload the atomic after the upgrade, or inline the load directly in the log:
```cpp
LOG_ERROR_CC(m_connection_context, "Maximum packet size exceed!, m_max_packet_size = "
  << m_max_packet_size.load(std::memory_order_relaxed)
  << ", packet header received " << m_current_head.m_cb
  << ", connection will be closed.");
```

contrib/epee/include/net/levin_base.h
R0BC0D3R: Fixed in 28a4b8b: wrapped 256*1024 in parentheses. Only one usage site exists (constructor initializer), so no functional change, but agreed it's the right defensive style for a macro.

contrib/epee/include/net/levin_protocol_handler_async.h
R0BC0D3R: Fixed in 28a4b8b: changed the LOG_ERROR_CC to call m_max_packet_size.load(std::memory_order_relaxed) directly instead of using the max_packet_size local captured before the handshake upgrade on line 586.

sn1f3rt (approved):
Both issues addressed correctly — macro parenthesized and log now reloads the atomic. LGTM.

sn1f3rt (approved):
Both issues addressed correctly — macro parenthesized and log now reloads the atomic. LGTM.

PR #83, opened 2026-05-15
fix: RPC denial-of-service mitigations

R0BC0D3R:
## Summary

Backports three upstream Monero security fixes targeting RPC denial-of-service
attack vectors:

**tx_extra size limit in mempool** (PR #8784, v0.18.2.2)
- Rejects mempool transactions with tx_extra exceeding 1060 bytes
- Prevents RAM/disk exhaustion via crafted oversized extra data
- Exempt for kept_by_block transactions to preserve sync compatibility
- RPC minor version bumped to 3.1, new `tx_extra_too_big` response field

**get_output_distribution DoS fix** (PR #8084, v0.17.3.0)
- Restricted RPC now rejects output distribution queries for pre-RCT
  (non-zero amount) outputs
- Prevents expensive enumeration queries from exhausting public node resources
- Adds missing CORE_RPC_ERROR_CODE_RESTRICTED (-14) error code

**RPC input size limit** (PR #8276, v0.17.3.2)
- All RPC endpoints now reject HTTP requests exceeding 1 MB
- Limit enforced at the epee HTTP handler level before any parsing occurs
- Tested: oversized requests are dropped with no response

## Test plan
- [ ] Normal daemon operation, sync, and transaction send unaffected
- [ ] RPC version reports 3.1
- [ ] Oversized RPC input returns empty response (connection dropped)
- [ ] get_output_distribution for non-RCT amounts rejected on restricted RPC

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -1191,6 +1191,8 @@ namespace cryptonote
         add_reason(reason, "tx is not ringct");
       if ((res.too_few_outputs = tvc.m_too_few_outputs))
         add_reason(reason, "too few outputs");
+      if ((res.tx_extra_too_big = tvc.m_tx_extra_too_big))
+        add_reason(reason, "tx-extra too big");
       const std::string punctuation = reason.empty() ? "" : ": ";
       if (tvc.m_verifivation_failed)
       {
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -240,6 +240,8 @@ namespace cryptonote
     if (rpc_config->login)
       http_login.emplace(std::move(rpc_config->login->username), std::move(rpc_config->login->password).password());
 
+    m_net_server.get_config_object().m_max_content_length = MAX_RPC_CONTENT_LENGTH;
+
     auto rng = [](size_t len, uint8_t *ptr){ return crypto::rand(len, ptr); };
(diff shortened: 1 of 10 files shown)


ngeojiajun (approved):

src/rpc/core_rpc_server.cpp
```diff
@@ -240,6 +240,8 @@ namespace cryptonote
     if (rpc_config->login)
       http_login.emplace(std::move(rpc_config->login->username), std::move(rpc_config->login->password).password());
 
+    m_net_server.get_config_object().m_max_content_length = MAX_RPC_CONTENT_LENGTH;
```
sn1f3rt, line 243: The 1 MB cap is applied here for the daemon RPC server, but `wallet_rpc_server::init()` (`src/wallet/wallet_rpc_server.cpp`) never sets `m_max_content_length`, so the wallet RPC server remains unprotected against oversized-request exhaustion. Consider setting the same limit there, or at least documenting the omission as intentional.

R0BC0D3R: Fixed in a401f01: added m_net_server.get_config_object().m_max_content_length = MAX_RPC_CONTENT_LENGTH; to wallet_rpc_server::init(), mirroring the daemon RPC fix.

sn1f3rt (approved):
All three fixes are correctly implemented and the wallet RPC gap has been addressed. Good to merge.

PR #84, opened 2026-05-15
fix: LMDB deadlock in resize detection during concurrent transaction begin

R0BC0D3R:
## Summary

- Fixes a low-probability deadlock in `lmdb_txn_begin()` triggered when
  `MDB_MAP_RESIZED` is returned during concurrent database access
- `lmdb_resized()` called `wait_no_active_txns()` without accounting for
  the active transaction held by the calling thread, causing it to spin
  forever waiting for a count that could never reach zero
- Fix passes `isactive` flag from call sites so the calling thread's own
  transaction is temporarily excluded from the active count during the wait

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -407,7 +407,12 @@ void mdb_txn_safe::allow_new_txns()
   creation_gate.clear();
 }
 
-void lmdb_resized(MDB_env *env)
+void mdb_txn_safe::increment_txns(int i)
+{
+  num_active_txns += i;
+}
+
+void lmdb_resized(MDB_env *env, int isactive)
 {
   mdb_txn_safe::prevent_new_txns();
 
@@ -418,7 +423,11 @@ void lmdb_resized(MDB_env *env)
   mdb_env_info(env, &mei);
   uint64_t old = mei.me_mapsize;
 
+  if (isactive)
+    mdb_txn_safe::increment_txns(-1);
   mdb_txn_safe::wait_no_active_txns();
+  if (isactive)
+    mdb_txn_safe::increment_txns(1);
 
   int result = mdb_env_set_mapsize(env, 0);
   if (result)
@@ -436,7 +445,7 @@ inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB
 {
   int res = mdb_txn_begin(env, parent, flags, txn);
   if (res == MDB_MAP_RESIZED) {
-    lmdb_resized(env);
+    lmdb_resized(env, 1);
     res = mdb_txn_begin(env, parent, flags, txn);
   }
   return res;
@@ -446,7 +455,7 @@ inline int lmdb_txn_renew(MDB_txn *txn)
 {
   int res = mdb_txn_renew(txn);
   if (res == MDB_MAP_RESIZED) {
-    lmdb_resized(mdb_txn_env(txn));
+    lmdb_resized(mdb_txn_env(txn), 0);
     res = mdb_txn_renew(txn);
   }
   return res;
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -441,7 +441,7 @@ void lmdb_resized(MDB_env *env, int is_active)
   mdb_txn_safe::allow_new_txns();
 }
 
-inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB_txn **txn, int is_active = 1)
+inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB_txn **txn, int is_active)
 {
   int res = mdb_txn_begin(env, parent, flags, txn);
   if (res == MDB_MAP_RESIZED) {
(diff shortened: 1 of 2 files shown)


ngeojiajun (approved):

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -418,7 +423,11 @@ void lmdb_resized(MDB_env *env)
   mdb_env_info(env, &mei);
   uint64_t old = mei.me_mapsize;
 
+  if (isactive)
+    mdb_txn_safe::increment_txns(-1);
   mdb_txn_safe::wait_no_active_txns();
+  if (isactive)
+    mdb_txn_safe::increment_txns(1);
 
   int result = mdb_env_set_mapsize(env, 0);
```
sn1f3rt, line 432: **Gap: `isactive=1` is incorrect for raw `MDB_txn*` call sites.**

`lmdb_txn_begin` is used in two distinct patterns:

1. With a `mdb_txn_safe` in scope (lines 1534, 3848, 4040, `TXN_PREFIX` macro) — `num_active_txns` was already incremented by the constructor, so `isactive=1` is correct here.
2. With a bare `MDB_txn*` and no surrounding `mdb_txn_safe` — `num_active_txns` was **never** incremented for this thread. Affected call sites: `build_block_cache` (line ~2406), `get_cna_v3_data` (line ~2502), and `block_rtxn_start` when invoked directly without `TXN_PREFIX_RDONLY` (line ~656 → line ~3991).

For pattern 2, `increment_txns(-1)` subtracts from a counter this thread never incremented:
- If `num_active_txns == 0`: the `uint64_t` wraps to `UINT64_MAX` and `wait_no_active_txns()` never exits — permanent deadlock, worse than the original bug.
- If `num_active_txns > 0`: incorrectly cancels out another live thread's transaction, potentially allowing the resize to proceed while that transaction is still active — LMDB corruption risk.

`isactive` needs to be passed **per call site** rather than hardcoded at the function level, since only the caller knows whether a `mdb_txn_safe` is currently in scope.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -436,7 +445,7 @@ inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB
 {
   int res = mdb_txn_begin(env, parent, flags, txn);
   if (res == MDB_MAP_RESIZED) {
-    lmdb_resized(env);
+    lmdb_resized(env, 1);
```
sn1f3rt: **Gap: `isactive=0` leaves the deadlock unfixed for all `TXN_PREFIX_RDONLY` read paths.**

`lmdb_txn_renew` is called exclusively from `block_rtxn_start`, which is reached via two paths:
1. The `TXN_PREFIX_RDONLY` macro (~35 read operations) — constructs `mdb_txn_safe auto_txn` **before** calling `block_rtxn_start`, so `num_active_txns` IS incremented for this thread.
2. A direct call at line ~656, with no `mdb_txn_safe` in scope.

In the `TXN_PREFIX_RDONLY` case, `block_rtxn_start` returns `true` (renew path), so `auto_txn.uncheck()` is never called and `num_active_txns` stays incremented. With `isactive=0`, the calling thread's own transaction is not excluded from the wait — if this is the last active thread, `wait_no_active_txns()` never returns. The original deadlock is unfixed for all read operations.

Same root cause as the `lmdb_txn_begin` comment above: `isactive` needs to be propagated from the call site. One approach: have `block_rtxn_start` accept a bool parameter indicating whether the caller has a live `mdb_txn_safe`, then pass the correct value through to `lmdb_txn_renew`.

src/blockchain_db/lmdb/db_lmdb.cpp
R0BC0D3R: Fixed in 1903c47. lmdb_txn_begin now takes an is_active parameter (default 1 for existing safe callers). build_block_cache and get_cna_v3_data are Nerva-specific functions that use bare MDB_txn* without a mdb_txn_safe wrapper. Both now explicitly pass is_active=0.

src/blockchain_db/lmdb/db_lmdb.cpp
R0BC0D3R: Fixed in 1903c47. lmdb_txn_renew and block_rtxn_start now both accept is_active propagated from the call site. TXN_PREFIX_RDONLY has mdb_txn_safe auto_txn in scope before calling block_rtxn_start, so it passes is_active=1. The direct call at the other block_rtxn_start site has no wrapper in scope and passes is_active=0.

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -432,21 +441,21 @@ void lmdb_resized(MDB_env *env)
   mdb_txn_safe::allow_new_txns();
 }
 
-inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB_txn **txn)
+inline int lmdb_txn_begin(MDB_env *env, MDB_txn *parent, unsigned int flags, MDB_txn **txn, int is_active = 1)
 {
```
sn1f3rt, line 445: Nit: defaulting `is_active` to `1` here is a footgun for future raw `MDB_txn*` callers who forget to pass `0` — they'd silently get the wrong behaviour. `lmdb_txn_renew` avoids this by requiring the argument explicitly (no default), which is the safer pattern. Consider dropping the default here too, or at least documenting the invariant (caller must have a live `mdb_txn_safe` in scope for `is_active=1` to be correct).

sn1f3rt (approved):
Both issues from the previous review are fully addressed. The `is_active` flag is now correctly propagated per call site across all paths. One nit inline.

src/blockchain_db/lmdb/db_lmdb.cpp
R0BC0D3R: Removed the default value entirely so every call site must declare is_active explicitly. The compiler now enforces correctness for any future caller.

ngeojiajun (approved):

PR #85, opened 2026-05-17
fix: wallet privacy hardening - decoy selection bias, DNS leaks, ring data attack surface

R0BC0D3R:
## Summary

Ported seven upstream Monero security and privacy fixes. All are NO-FORK and independent of each other.

### Decoy selection fixes (wallet2.cpp)

- **Deprecate `find_and_save_rings()`** — removes attack surface where a
  hostile RPC node could exploit the wallet scanning path to extract ring
  data. Rings are already saved during scanning since the last Nerva HF;
  the function was dead code. (PR #10015)

- **Integer truncation in `average_output_time`** — integer division
  truncated fractional values (e.g. 1.9 → 1), skewing decoy output age
  selection and biasing anonymity sets. (PR #7798)

- **Gamma distribution from chain tip** — the gamma picker was computing
  output age relative to the unlock time rather than the chain tip,
  producing biased decoy selection as identified in Miller et al.
  (https://arxiv.org/pdf/1704.04299/). (PR #7822)

- **Recent spend window 50 → 15 blocks** — after fixing integer
  truncation, the original 50-block window was oversized. 15 better
  matches empirically observed spending patterns. (PR #7993)

- **Statistical dependence within rings** — sequential picking without
  replacement creates bias between later picks. Introduces
  `secret_picking_order` to commit to pick ordering before RPC
  validation, so earlier (less biased) picks are preferred. (PR #9130)

### DNS privacy leak fixes

- **Remove DNS calls from `get_segregation_fork_height()`** — dead code
  queried `segheights.moneropulse.org` on every tx construction path.
  DNS URLs were already cleared in Nerva but the code structure remained.
  (PR #8408)

- **Replace DNS resolver in `is_local_address()`** — wallet performed a
  DNS lookup to determine whether the daemon is local, leaking the
  connected node address on every startup. Replaced with direct IP
  parsing via `boost::asio::ip::make_address()` and RFC 6761 localhost
  matching. (PR #8878)

## Notes

- Unit tests from PR #8878 omitted — Nerva lacks that test infrastructure
- `DIFFICULTY_TARGET_V2` references in upstream replaced with
  `DIFFICULTY_TARGET` throughout (Nerva uses a single 60s target)


sn1f3rt:
The `RECENT_SPEND_WINDOW` value of 15 blocks was empirically justified in Monero PR #7993 using chain data at 120-second blocks (= 1800 seconds). Nerva runs 60-second blocks, so 15 blocks = **900 seconds** here — half the window. No equivalent analysis has been done on Nerva chain data, so it's unclear whether 15 blocks is still the right fit for observed spending patterns on this chain.

@ngeojiajun — thoughts?

ngeojiajun:
Agreed, since we do not have clear idea on spending pattern lets at least make it similar to monero in term of time. So i think we should use 30 blocks instead

sn1f3rt:
> Agreed, since we do not have clear idea on spending pattern lets at least make it similar to monero in term of time. So i think we should use 30 blocks instead

+1, @R0BC0D3R for viz

R0BC0D3R:
Changed from 15 to 30 blocks. Monero's 15 × 120s = 1800s; Nerva's 30 × 60s = 1800s, preserving the same 30-minute real-time window.

sn1f3rt (approved):
LGTM

PR #86, opened 2026-05-17
fix: wallet correctness — pool spend detection and key image state

R0BC0D3R:
## Summary

- **Pool spend detection during scanning** — When refreshing, the wallet
  now detects outgoing transactions in the mempool even when the wallet
  didn't originate them (e.g. after restoring from seed or running a
  parallel wallet with the same seed). Outputs are marked spent
  immediately when a matching key image is found in a pool transaction,
  and the tx is added to the unconfirmed list. First refresh also clears
  the pool scan cache so restored wallets get a complete view of pending
  spends.

- **Key image spent status correctness** — When a pending transaction
  re-appears in the pool after previously being marked failed, its
  outputs are now correctly re-marked as spent. The output unspent reset
  on tx failure was also refactored to use `m_key_images` O(1) lookup
  instead of an O(n) linear scan, with an idempotency guard to skip
  outputs already in the correct state.

## Test plan

- [ ] Send a transaction via wallet A; refresh a parallel wallet B
  (same seed, different file) while tx is in pool — verify it appears
  in `show_transfers pending` on wallet B
- [ ] Confirm balance reflects pool spend immediately after refresh,
  not only after block confirmation
- [ ] Verify normal send/receive flow unaffected

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -1550,16 +1550,11 @@ void wallet2::set_unspent(size_t idx)
   td.m_spent_height = 0;
 }
 //----------------------------------------------------------------------------------------------------
-bool wallet2::is_spent(const transfer_details &td, bool strict) const
+bool wallet2::is_spent(const transfer_details &td, bool /*strict*/) const
 {
-  if (strict)
-  {
-    return td.m_spent && td.m_spent_height > 0;
-  }
-  else
-  {
-    return td.m_spent;
-  }
+  // Pool-spent outputs (m_spent_height==0) are still spent; the strict flag
+  // only controls whether unconfirmed change is added back in balance_per_subaddress.
+  return td.m_spent;
 }
 //----------------------------------------------------------------------------------------------------
 bool wallet2::is_spent(size_t idx, bool strict) const
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2228,7 +2228,11 @@ void wallet2::process_new_transaction(const crypto::hash &txid, const cryptonote
     else if (!m_unconfirmed_txs.count(txid))
     {
       // Add to unconfirmed txs if not already there (e.g. restoring wallet, or running the wallet in parallel to the sending wallet w/same seed)
-      add_unconfirmed_tx(txid, tx, tx_money_spent_in_ins, {}/*don't know dests*/, crypto::null_hash/*don't know payment_id*/, self_received, *subaddr_account, subaddr_indices);
+      crypto::hash payment_id = crypto::null_hash;
+      tx_extra_nonce extra_nonce;
+      if (find_tx_extra_field_by_type(tx_extra_fields, extra_nonce))
+        get_payment_id_from_tx_extra_nonce(extra_nonce.nonce, payment_id);
+      add_unconfirmed_tx(txid, tx, tx_money_spent_in_ins, {}/*don't know dests*/, payment_id, self_received, *subaddr_account, subaddr_indices);
       auto i = m_unconfirmed_txs.find(txid);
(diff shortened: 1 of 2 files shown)


sn1f3rt (commented):
Found three bugs after reviewing the changes:

**Bug 1 — `set_spent(height=0)` breaks `is_spent(strict=true)`**

`set_spent(it->second, height)` is now called unconditionally, but `process_pool_state` always passes `height=0`:

```cpp
process_new_transaction(tx_hash, tx, std::vector<uint64_t>(), 0, 0, now, false, true, ...);
```

`is_spent` with `strict=true` checks `td.m_spent && td.m_spent_height > 0`, so a pool-spent output (`m_spent=true, m_spent_height=0`) is treated as unspent in strict mode. This causes `balance(strict=true)` to double-count pool-spent outputs and `get_reserve_proof` (which hardcodes `strict=true` at line 10853) to include them in the proof.

Fix: use a non-zero sentinel for the unconfirmed height, or add a separate `m_pool_spent` flag to `transfer_details`.

---

**Bug 2 — `m_state` not reset when a failed tx reappears in the pool**

The new `else` branch in `update_pool_state` correctly re-marks outputs as spent when a previously-failed tx reappears in the pool, but never updates `m_state`:

```cpp
else
{
    // re-mark outputs as spent
    set_spent(i, 0);
    // m_state stays `failed`
}
```

`balance_per_subaddress` filters out failed-state entries from change accounting (`m_state != failed`), so the balance reflects neither the spend nor the change even though the outputs are gone. `show_transfers` also shows the tx as failed while outputs are marked spent.

Fix: add `pit->second.m_state = wallet2::unconfirmed_transfer_details::pending;` in the `else` branch.

---

**Bug 3 — Empty `m_dests` and null payment ID propagate into confirmed tx history**

`add_unconfirmed_tx` is called with `{}` for dests and `crypto::null_hash` for payment ID. When the tx confirms, `process_unconfirmed` constructs `confirmed_transfer_details` directly from the unconfirmed entry (copying the empty `m_dests`), and `process_outgoing` finds `entry.second = false` so skips the block that would populate `m_dests`. The confirmed tx history permanently shows no destinations or payment ID for txs detected via this path.

This is partially unavoidable without the tx keys, but should be documented as a known limitation.

R0BC0D3R:
Fixed in f4d3a6e. Gated the set_spent call on m_state == failed to prevent unconditional re-marking on every pool sync, reset m_state to pending when the tx reappears so balance_per_subaddress correctly includes change, and extracted the unencrypted payment_id from tx_extra_fields (already in scope) instead of passing null_hash to add_unconfirmed_tx.

sn1f3rt:
Re-reviewed after the latest push.

**Bug 2** — Fixed. The `else` branch is now correctly gated on `m_state == failed` and resets the state to `pending` before re-marking outputs as spent.

**Bug 3** — Partially addressed. Payment ID is now extracted from `tx_extra` rather than hardcoded as `null_hash`, which is the right call. However `m_dests` is still passed as `{}`, so destination addresses will still be missing from confirmed tx history for txs detected via this path. Since those can't be recovered without tx keys this is unavoidable, but worth a comment in the code so it's clear this is a known limitation rather than an oversight.

**Bug 1** — Still present. `set_spent(it->second, height)` remains outside the `!pool` guard and `process_pool_state` still calls `process_new_transaction` with `height=0`, so pool-spent outputs end up with `m_spent=true, m_spent_height=0`. `is_spent(strict=true)` returns false for those (`td.m_spent && td.m_spent_height > 0`), meaning `balance(strict=true)` and `get_reserve_proof` will still incorrectly include them.

R0BC0D3R:
Fixed in 77f4937: Removed the m_spent_height > 0 requirement from is_spent(strict=true). Pool-spent outputs now correctly appear as spent in all modes. The strict parameter is kept in the API since balance_per_subaddress still uses it to control whether unconfirmed change is added back.

sn1f3rt (approved):
LGTM

PR #87, opened 2026-05-18
fix: fall back to single DNS seed record when only one domain responds

R0BC0D3R:
## Summary

The dual-validation check in `load_txt_records_from_dns` requires TXT records from two domains to match before accepting any seed nodes. The single-domain shortcut that bypassed this check was gated on `dns_urls.size() == 1` — the number of domains *configured* — rather than the number of domains that actually *returned* records.

With two domains configured, if one goes silent (domain expiry, DNSSEC failure, missing TXT record), `num_valid_records` drops to 1 but `dns_urls.size()` remains 2, so the shortcut never fires. The result is no seed nodes returned and a `WARNING: no two DNS TXT records matched` log entry, even though one perfectly valid DNSSEC-verified record is available.

## Change

```cpp
// before
if (dns_urls.size() == 1)

// after
if (num_valid_records == 1)

### DIFF
--- a/src/common/dns_utils.cpp
+++ b/src/common/dns_utils.cpp
@@ -566,7 +566,7 @@ bool load_txt_records_from_dns(DNSResolver &dr, std::vector<std::string> &good_r
   //nerva currently only has one dns update url. So if we have made it this far
   //we have a dnssec verified update record, so accept it. it is after all only for notification purposes
   //the code will automatically require 2 if it comes a time when we can add a second domain
-  if (dns_urls.size() == 1)
+  if (num_valid_records == 1)
   {
     good_records = records[0];
     return true;
--- a/src/common/dns_utils.cpp
+++ b/src/common/dns_utils.cpp
@@ -563,13 +563,21 @@ bool load_txt_records_from_dns(DNSResolver &dr, std::vector<std::string> &good_r
     return false;
   }
 
-  //nerva currently only has one dns update url. So if we have made it this far
-  //we have a dnssec verified update record, so accept it. it is after all only for notification purposes
-  //the code will automatically require 2 if it comes a time when we can add a second domain
+  // Only one domain responded. Cross-validation is skipped, but the record is
+  // still DNSSEC-verified. Warn only for multi-domain configs where one domain
+  // went silent — an attacker suppressing one domain could exploit this path.
   if (num_valid_records == 1)
   {
-    good_records = records[0];
-    return true;
+    if (records.size() > 1)
+      LOG_PRINT_L0("WARNING: only one DNS domain responded; using single DNSSEC-verified record without cross-validation");
+    for (const auto& record_set : records)
+    {
+      if (!record_set.empty())
+      {
+        good_records = record_set;
+        return true;
+      }
+    }
   }
   int good_records_index = -1;
   for (size_t i = 0; i < records.size() - 1; ++i)

### CODE REVIEW

File: src/common/dns_utils.cpp [hunk 1]
```diff
@@ -566,7 +566,7 @@ bool load_txt_records_from_dns(DNSResolver &dr, std::vector<std::string> &good_r
   //nerva currently only has one dns update url. So if we have made it this far
   //we have a dnssec verified update record, so accept it. it is after all only for notification purposes
   //the code will automatically require 2 if it comes a time when we can add a second domain
-  if (dns_urls.size() == 1)
+  if (num_valid_records == 1)
   {
     good_records = records[0];
```
sn1f3rt: **[CRITICAL] Wrong index used — returns empty `good_records` when the first domain is the failing one**

The body still reads `good_records = records[0]` after the condition changes. With `TXT_SEED_NODES` having two domains (`seed.nerva.one` at index 0, `seed.nerva.info` at index 1): if `dns_urls[0]` fails DNSSEC its record is cleared (lines 534/539), leaving `records[0]` empty. When `num_valid_records == 1` fires, `good_records` is set to an empty vector and the function returns `true` — the caller sees success but gets no seed nodes.

The fix only works when the *second* domain fails. It silently breaks the case it was written for when the *first* domain is the failing one.

Suggested fix — iterate to find the non-empty record:
```cpp
if (num_valid_records == 1)
{
    for (size_t i = 0; i < records.size(); ++i)
    {
        if (!records[i].empty())
        {
            good_records = records[i];
            return true;
        }
    }
}
```

sn1f3rt (commented):
Review of the fallback logic change. Four issues found — one critical blocker, two major concerns, one minor.

src/common/dns_utils.cpp
```diff
@@ -566,7 +566,7 @@ bool load_txt_records_from_dns(DNSResolver &dr, std::vector<std::string> &good_r
   //nerva currently only has one dns update url. So if we have made it this far
   //we have a dnssec verified update record, so accept it. it is after all only for notification purposes
   //the code will automatically require 2 if it comes a time when we can add a second domain
-  if (dns_urls.size() == 1)
+  if (num_valid_records == 1)
   {
```
sn1f3rt, line 570: **[MAJOR] Security regression: bypassing cross-validation weakens seed node poisoning protection**

`TXT_SEED_NODES` is the only two-domain config (both mainnet and testnet). The dual-domain requirement — with `dns_records_match` cross-validation at lines 575–596 — is the anti-DNS-poisoning mechanism for seed discovery. Both domains must agree before their records are trusted.

This change bypasses that check whenever one domain is silent. DNS silence is itself an exploitable condition: an attacker who can suppress responses from one domain (BGP hijack, targeted DNS interception) can force the single-record fallback and inject arbitrary seed nodes through the remaining unvalidated domain. The PR description frames this purely as a reliability fix but doesn't acknowledge the security tradeoff.

sn1f3rt, line 570: **[MAJOR] Semantic contract broken for future multi-domain callers**

The old `dns_urls.size() == 1` guard was structurally meaningful: it explicitly said "only skip dual-validation for single-domain configs." The new `num_valid_records == 1` fires for any multi-domain config the moment one domain goes quiet.

For the current callers this is correct by construction (single-domain configs always have `num_valid_records == 1` when they pass), but the structural guarantee is gone. Any future two-domain caller that expects mandatory cross-validation will silently get single-record fallback whenever one domain is unresponsive, with no indication the check was skipped.

src/common/dns_utils.cpp
```diff
@@ -566,7 +566,7 @@ bool load_txt_records_from_dns(DNSResolver &dr, std::vector<std::string> &good_r
   //nerva currently only has one dns update url. So if we have made it this far
   //we have a dnssec verified update record, so accept it. it is after all only for notification purposes
```
sn1f3rt: **[MINOR] Stale comment no longer describes the condition**

This comment was written to explain the `dns_urls.size() == 1` guard — it says "nerva currently only has one dns update url" and "the code will automatically require 2 if it comes a time when we can add a second domain." Both statements are now factually wrong: there are already two seed node domains, and the condition changed to `num_valid_records == 1`. The comment should be updated to reflect what the new condition actually does and why.

src/common/dns_utils.cpp
R0BC0D3R: Fixed in b8bcff6. Replaced records[0] with a loop that iterates to find the first non-empty record set, so the correct result is returned regardless of which domain failed DNSSEC validation.

src/common/dns_utils.cpp
R0BC0D3R: Acknowledged. The cross-validation bypass is a real tradeoff. The remaining domain is still DNSSEC-verified, so an attacker would also need to compromise DNSSEC on that domain, not just suppress the other. Added a WARNING log so the fallback is visible to operators rather than silent.

R0BC0D3R: The warning log makes the degraded state explicit rather than silent. A stricter fix (returning false when any configured domain fails) would reintroduce the original bug this PR was created to solve. Accepted as a documented tradeoff.

src/common/dns_utils.cpp
R0BC0D3R: Fixed in b8bcff6: Updated the comment to accurately describe what the num_valid_records == 1 path does and why, replacing the outdated reference to a single-URL configuration.

sn1f3rt (approved):
LGTM

PR #90, opened 2026-05-24
fix: prevent crash in build_block_cache on Windows (LMDB write txn reuse)

R0BC0D3R:
## Problem

nervad crashes after running for some time on Windows when fully synced
and mining. The crash is preceded by a single stderr line:

    get_v3_data failed, error 22 Invalid argument

and then an immediate access violation.

**Root cause:** `build_block_cache` was reusing the active batch write
transaction (`m_write_txn`) when `m_batch_active` was set. On Windows
with `MDB_WRITEMAP`, `mdb_cursor_open` returns `EINVAL` (22) when given
that write txn in certain states. The previous `check_error` only printed
to stderr and returned, leaving the cursor pointer uninitialised. The
subsequent `mdb_cursor_get` call dereferenced the garbage pointer, causing
an access violation and crash.

## Fix

- `build_block_cache` now always opens a fresh read-only transaction
  instead of reusing the write txn.
- `check_error` now throws `DB_ERROR` instead of printing to stderr, so
  any future LMDB failures surface as catchable exceptions rather than
  silent continuation with an invalid cursor.

The write txn reuse was never necessary: `get_block_sync_size()` caps
batch size at `BLOCKS_SYNCHRONIZING_SAFE_BATCH_COUNT` (256), which
matches the `stable_height = height-256` offset used by all CNA PoW
variants. The blocks needed by `build_block_cache` are therefore always
in a prior committed batch and visible to a read-only transaction.

## Testing

Confirmed on Windows 11 with nervad v0.2.2.0-rc1 (7746fa8b). Previously
crashed multiple times per day; ran for over 20 hours continuously with
this fix applied and zero LMDB errors in the log.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -2385,7 +2385,7 @@ block_header BlockchainLMDB::get_block_header(const crypto::hash& h) const
 void check_error(int err)
 {
   if (err)
-    fprintf(stderr, "get_v3_data failed, error %d %s\n", err, mdb_strerror(err));
+    throw0(DB_ERROR(lmdb_error("LMDB cursor error in PoW cache: ", err).c_str()));
 }
 
 void BlockchainLMDB::build_block_cache(uint64_t height)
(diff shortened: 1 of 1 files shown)


sn1f3rt (approved):
LGTM

ngeojiajun (approved):

PR #92, opened 2026-06-05
feat: enable hardware AES on ARMv8

mmokhi:
## Summary

ARMv8 builds shipped with software AES even on boards that support the Crypto Extensions. This adds a NEON Cryptonight AES path and dispatches between HW and SW once at startup, so the same binary works on AES-capable and AES-less ARMv8 CPUs.

## What changed

- NEON Cryptonight path using `vaeseq_u8` + `vaesmcq_u8`. `_mm_aesenc_si128` is polyfilled to `veorq_u8(vaesmcq_u8(vaeseq_u8(a, 0)), k)`, same semantics as x86 `aesenc`.
- `aes_pseudo_round` / `aes_pseudo_round_xor` have NEON-native implementations.
- `aes_expand_key` on ARM defers to `oaes_key_import_data` so the expanded key matches the SW path exactly. I tried Monero's inline-asm AES-256 schedule first but the self-test caught a mismatch (that asm has never been compiled into a shipped Monero binary because their depends also forces `NO_AES` on ARM).
- `slow-hash.c` is split into a dispatcher, `slow-hash-hw.c`, and `slow-hash-sw.c`. Runtime dispatch via `getauxval(AT_HWCAP) & HWCAP_AES` on Linux/Android, `__builtin_cpu_supports("aes")` on x86, true on Apple Silicon. `NERVA_FORCE_SOFTWARE_AES=1` in the env forces SW.
- `slow-hash-hw.c` is compiled with `-march=armv8-a+crypto` per file. The rest of the binary stays at baseline `armv8-a` so it still loads on AES-less hardware.
- `check_aesni()` runs a HW-vs-SW self-test at startup and refuses to start if the two paths disagree.
- `contrib/depends/toolchain.cmake.in` no longer forces `NO_AES` on aarch64 (linux, android, darwin). armv7 is unchanged, no Crypto Extensions there.

## Test plan

- [x] CI cross-build matrix green for ARM v8, ARM v7, x86_64 Linux, i686 Linux, Win64, i686 Win, Cross-Mac x86_64, Cross-Mac aarch64, x86_64 FreeBSD.
- [x] Verified on a Jetson Nano 2GB (Cortex-A57 with Crypto Extensions): "Hardware AES detected and self-test passed" at startup, genesis block adds cleanly, daemon runs.
- [x] SW fallback exercised by setting `NERVA_FORCE_SOFTWARE_AES=1`.
- [x] Also tested on AES-less ARMv8 hardware (Cortex-A72 / RPi 4). The dispatcher logic and SW path are the same code that ships in the current `NO_AES`-forced ARMv8 builds, so it should be fine, but was worth confirming.


sn1f3rt (commented):
Reviewed the ARMv8 hardware-AES path. Overall a careful, well-structured change: the NEON AES reformulation is mathematically equivalent to x86 `aesenc`, the key schedules match via `oaes`, the x86 path is a verbatim code move, and the runtime fail-closed self-test is the right safety mechanism. Findings below in roughly descending importance. The one I'd treat as should-fix-before-release is the self-test coverage gap — it validates `v7_8` while the live PoW is `v11`.

src/crypto/slow-hash.c
```diff
@@ -131,377 +235,29 @@ void cn_hash_context_free(cn_hash_context_t *context)
...
-    randomize_scratchpad(context->random_values, hp_state);
-    xor_u64();
-
-    for (i = 0; i < iters; i++) {
-        aes_sw_variant();
-    }
-
-    finalize_hash();
-}
+    cn_slow_hash_v7_8_hw(ctx, input, sizeof(input) - 1, hw, 64);
```
sn1f3rt, line 247: **Self-test validates `v7_8`, but the live PoW is `v11`.** `get_block_longhash` routes the current hardfork (the `default:` case in `cryptonote_tx_utils.cpp`) to `cn_slow_hash_v11`. The self-test never exercises v11/v10, which contain HW-vs-SW logic that v7_8 lacks — notably the post-AES state read feeding `salt_pad`, which uses a different buffer on each path: HW `r2 = (uint16_t*)&c` vs SW `r2 = (uint16_t*)&b` in `slow-hash-impl.h`.

That cross-path equivalence is load-bearing and is currently true on mainnet (x86-HW `&c` and ARM-SW `&b` coexist without splitting), so switching ARM-crypto nodes to HW-v11 is most likely safe. But the fail-closed self-test — the whole safety mechanism here — doesn't cover the variant that actually secures the chain. If an ARM-crypto-only divergence ever existed in the v10/v11 path, the daemon would start cleanly and then compute wrong PoW.

Recommend extending the self-test to also compare `cn_slow_hash_v11` (and ideally `v10`) with fixed `xx`/`yy`/`init_size_blk`. `cn_hash_context_create` already allocates `ctx->salt`, and both paths read the same context buffers, so the comparison stays deterministic.

src/common/util.cpp
```diff
@@ -798,12 +799,24 @@ std::string get_nix_version_display_string()
   bool check_aesni()
   {
 #if !defined NO_AES
-    if (!crypto::has_aesni())
+    if (crypto::cn_hardware_aes_supported())
```
sn1f3rt, line 802: **Behavior change worth confirming.** Previously `check_aesni()` hard-failed on x86 without AES-NI (`!has_aesni()` → `MGUSER_RED` + `return false`). Now it warns and continues on the software path. Reasonable improvement, but please confirm we want an AES-less / misconfigured x86 node to silently run (very slow) software hashing rather than refuse to start. Note the binary is still globally `-maes`; the SW fallback only survives because the compiler doesn't emit AES instructions outside the intrinsics in `slow-hash-hw.c`.

src/crypto/slow-hash.c
```diff
@@ -29,19 +29,127 @@
...
+{
+    const char *force_sw = getenv("NERVA_FORCE_SOFTWARE_AES");
+    if (force_sw != NULL && force_sw[0] != '\0' && force_sw[0] != '0')
+        return 0;
+
+#if !defined(SLOW_HASH_HW_AES_BUILT)
+    return 0;
+#elif defined(__APPLE__) && defined(__aarch64__)
+    /* Apple Silicon always implements the ARMv8 Crypto Extensions. */
+    return 1;
```
sn1f3rt, line 81: **Verify the cross-Mac arm64 build actually enables this branch.** This `return 1` is only reachable if `SLOW_HASH_HW_AES_BUILT` is defined, which requires `CN_ARM_CRYPTO_FLAG` to be set — i.e. `CHECK_CXX_ACCEPTS_FLAG("-march=armv8-a+crypto")` must pass on the Mac cross-toolchain. If that probe fails, `slow-hash-hw.c` isn't compiled, `detect_hardware_aes()` returns 0 at the first `#if`, and Apple Silicon **silently runs software AES** while this branch becomes dead code. CI being green only proves it compiled. Please confirm the cross-Mac aarch64 build log shows "Crypto extensions available for ARMv8". (Also: per the test plan, the HW path is currently untested on real Apple Silicon and on AES-less ARMv8 / RPi 4.)

src/crypto/slow-hash.h
```diff
@@ -306,6 +361,81 @@ STATIC INLINE void xor64(uint64_t *a, const uint64_t b)
...
+ * scratchpad init diverges and the final hash differs.
+ *
+ * We tried an inline-asm AES-256 schedule borrowed from Monero upstream
+ * (which forces NO_AES on ARM, so that asm has never actually run in
+ * production); the self-test in check_aesni() caught a mismatch. Using
+ * oaes is slower per call but the function only fires twice per cn_slow_hash,
+ * so the perf impact is negligible next to the 1MB scratchpad fill. */
+STATIC INLINE void aes_expand_key(const uint8_t *key, uint8_t *expandedKey)
+{
+    OAES_CTX *ctx = oaes_alloc();
```
sn1f3rt: Two notes on this function:

1. **Missing NULL check on `oaes_alloc()`.** On OOM, `oaes_key_import_data(NULL, …)` returns safely (`OAES_RET_ARG1`), but the next line `c->key->exp_data_len` dereferences the NULL `ctx` → crash. Add a guard.

2. **`oaes_alloc`/`oaes_free` per call (perf).** This fires twice per `cn_slow_hash` (via the `expand_key()` and `finalize_hash()` macros) — thousands of malloc/free pairs per second per core when mining, i.e. heap churn / allocator contention in the hot path. The context already owns a reusable `oaes_ctx`; threading `context` in here would avoid the per-call allocation. Correctness is fine — just an optimization.

src/crypto/CMakeLists.txt
```diff
@@ -72,7 +73,27 @@ set(crypto_private_headers
...
+# dispatcher in slow-hash.c picks _hw or _sw based on getauxval(AT_HWCAP)
+# on Linux/Android and the existing cpuid path on x86.
+set(_slow_hash_hw_supported FALSE)
+if (NOT NO_AES)
+  if (ARM8 AND CN_ARM_CRYPTO_FLAG)
+    set(_slow_hash_hw_supported TRUE)
+    list(APPEND crypto_sources slow-hash-hw.c)
+    set_source_files_properties(slow-hash-hw.c PROPERTIES
+      COMPILE_FLAGS "${CN_ARM_CRYPTO_FLAG}")
+  elseif (NOT ARM AND NOT PPC64LE AND NOT PPC64 AND NOT PPC AND NOT S390X AND NOT RISCV AND NOT LOONGARCH)
```
sn1f3rt: **i686 builds a redundant translation unit.** For 32-bit x86 this branch is taken (not ARM/PPC/…), so `slow-hash-hw.c` is compiled and `SLOW_HASH_HW_AES_BUILT` is defined. But in `slow-hash.h` the HW-AES guard requires `__x86_64__ || (_MSC_VER && _WIN64) || (__aarch64__ && __ARM_FEATURE_CRYPTO)` — i686 matches none, so `cn_slow_hash_hw` compiles as the *software* body, and `detect_hardware_aes()` returns 0 on i686 so it's never called. Correct, just a wasted duplicate TU. Consider gating this on a 64-bit/`__x86_64__` check to skip the HW TU on i686.

src/crypto/slow-hash.h
```diff
@@ -168,7 +179,46 @@ BOOL SetLockPagesPrivilege(HANDLE hProcess, BOOL bEnable)
     randomize_scratchpad(r, scratchpad);
 
 
-#if !defined(NO_AES) && (defined(__x86_64__) || (defined(_MSC_VER) && defined(_WIN64)))
+#if !defined(CN_FORCE_SOFTWARE_AES) && !defined(NO_AES) && (         \
```
sn1f3rt, line 182: Nit: the matching `#endif` for this guard (line 711) still carries the old trailing comment `!defined(NO_AES) && (defined(__x86_64__) || (defined(_MSC_VER) && defined(_WIN64)))`, which no longer reflects this `#if` now that it also covers `__aarch64__ && __ARM_FEATURE_CRYPTO` and `!defined(CN_FORCE_SOFTWARE_AES)`. Worth updating that closing comment to match.

src/common/util.cpp
mmokhi: Yes, intentional. The dispatcher now compiles both slow-hash-hw.c and
slow-hash-sw.c into the binary, so on a non-AES-NI x86 CPU the daemon
dispatches to the SW path and runs slow-but-correct hashing instead of
refusing to start. But if the community would rather we keep the hard-fail
behavior, happy to change it back. I tried to keep the fallback explicit
in the startup logs so it's not silent.

On the global -maes concern: aesb.c is pure-C table-based AES with no
intrinsics, so gcc only emits AES instructions where the intrinsics are
written (inside slow-hash-hw.c). On a CPU without AES-NI,
detect_hardware_aes() returns 0, the dispatcher never enters
slow-hash-hw.c, and the AES instructions in that TU never get executed.

src/crypto/slow-hash.c
mmokhi: Confirmed. The Cross-Mac aarch64 CI job's configure log shows:

    -- Checking to see if CXX compiler accepts flag -march=armv8-a+crypto - yes
    -- Crypto extensions available for ARMv8, hardware AES will be dispatched at runtime

So slow-hash-hw.c gets the +crypto flag and the always-true-on-Apple-Silicon
branch in detect_hardware_aes() is reachable.

src/crypto/slow-hash.c
mmokhi: Done in d6b3c19. Extended to cover v9, v10, v11 (v10/v11 with small xx/yy so salt_pad gets exercised). Each pair resets ctx->salt between HW and SW because v10/v11 write back via salt_pad.

src/crypto/slow-hash.h
mmokhi: Done in d6b3c19. aes_expand_key now takes an OAES_CTX* and borrows
ctx->oaes_ctx instead of doing oaes_alloc/oaes_free per call. This
saves one malloc/free pair per call (the outer ctx struct);
oaes_key_import_data still destroys and re-allocates the inner key
buffers internally, so heap pressure isn't eliminated entirely.
Removing the rest would need an oaes_lib API change to update keys
in place, out of scope here. NULL guard added so a NULL ctx zeros
the expansion instead of crashing.

src/crypto/CMakeLists.txt
mmokhi: Done in d6b3c19. Added BUILD_64 to the cmake conditional. i686 no longer compiles slow-hash-hw.c.

src/crypto/slow-hash.h
mmokhi: Done in d6b3c19. Updated to reflect the current guard.

src/crypto/slow-hash.c
sn1f3rt: Verified against d6b3c19 — v9/v10/v11 are now covered, the `_hw`/`_sw` symbol renames exist for all three, and the self-test call signatures line up with the function defs (v9: 5 args, v10: 10, v11: 8). The `salt` reset between each HW and SW call (and the `random_values` reset per variant) is correct, and with `xx=yy=2` the `salt_pad`/`r2` path is actually exercised — so the `&c`-vs-`&b` equivalence is now guarded at startup for the live v11 PoW. Exactly what was needed; thanks. Resolving.

src/common/util.cpp
sn1f3rt: Thanks for laying out the trade-off. Decision: keep the graceful software fallback as you have it. The fail-closed net still fires on the case that actually matters (HW hash != SW hash -> daemon aborts via `return 1` out of `main`), and a node with no AES-NI running slow-but-correct is preferable to locking it out — which also fits the broader goal of widening hardware support. No change needed. Resolving.

src/crypto/slow-hash.c
sn1f3rt: Confirmed by the configure-log excerpt (`-march=armv8-a+crypto - yes`, "Crypto extensions available for ARMv8") — the `+crypto` flag reaches `slow-hash-hw.c` and the Apple-Silicon `return 1` branch is live. One non-blocking item before we tag a release: the runtime path is still unverified on real Apple Silicon / AES-less ARMv8 hardware (CI only proves it compiles and the toolchain probe passes). The startup self-test is the safety net, but a single real arm64 run logging `cn_hardware_aes_supported() = 1` plus a clean self-test would close the loop. Resolving the thread; tracking the hardware run separately.

src/crypto/slow-hash.h
sn1f3rt: Verified in d6b3c19 — `aes_expand_key` now borrows `context->oaes_ctx`, dropping the per-call `oaes_alloc`/`free` pair. Correctness holds: the HW and SW paths are mutually exclusive per call, and on the HW path the borrowed ctx is only used transiently for key expansion (the rounds use NEON). The NULL guard is effectively defensive-only, since `cn_hash_context_create` returns NULL if `oaes_alloc` fails — so a live context always has a non-NULL `oaes_ctx`. Agreed that removing the inner `oaes_key_import_data` re-alloc would need an oaes API change and is out of scope here. Resolving.

src/crypto/CMakeLists.txt
sn1f3rt: Confirmed — `BUILD_64` added to the cmake conditional, so i686 no longer compiles the HW TU. Resolving.

src/crypto/slow-hash.h
sn1f3rt: Confirmed updated to match the current guard. Resolving.

sn1f3rt (approved):
**Approving for merge — the implementation is sound and all review feedback is addressed.**

Re-reviewed the latest revision (d6b3c19) against the code, not just the replies:

- **Self-test** now covers v7_8 / v9 / v10 / **v11** — i.e. the live mainnet PoW. Call signatures match the function defs, `random_values` is reset per variant, and `salt` is reset between each HW and SW call (v10/v11 write back via `salt_pad`). With `xx=yy=2` the `salt_pad`/`r2` path is actually exercised, so the HW-vs-SW equivalence — including the `r2 = &c` (HW) vs `&b` (SW) source-buffer difference — is guarded **fail-closed at startup**.
- **`aes_expand_key`** borrows the context's `oaes_ctx` instead of doing a per-call `oaes_alloc`/`free`, with a NULL guard (defensive-only, since `cn_hash_context_create` returns NULL if the alloc fails). HW and SW paths are mutually exclusive per call, so borrowing is safe.
- **x86 AES-NI detection** now reuses the project's tested `crypto::has_aesni()` CPUID path via a C wrapper — this also fixes the Cross-Mac x86_64 `___cpu_model` link error. The MinGW "multiple definition" of `SetLockPagesPrivilege` is fixed too.
- **i686** no longer compiles the HW TU (`BUILD_64` gate); stale `#endif` guard comment updated.
- The **x86 no-AES-NI behavior change** (graceful software fallback instead of hard-fail) is intentional and accepted — the fail-closed abort still fires on the case that matters (a HW path that produces a different hash than SW).

The NEON AES reformulation is byte-equivalent to x86 `aesenc`, the C/C++ linkage is correct, and the self-test is deterministic.

**Follow-up (not blocking merge):** the HW path hasn't yet been exercised on real Apple Silicon or AES-less ARMv8 (e.g. Raspberry Pi 4) hardware — CI only proves it compiles and that the `-march=armv8-a+crypto` probe passes. Worth capturing one real arm64 run logging `cn_hardware_aes_supported() = 1` with a passing startup self-test when someone has the hardware to hand. Safe to merge before then: the fail-closed self-test means any HW/SW divergence refuses to start rather than corrupting consensus.

mmokhi:
> **Approving — the implementation is sound and all review feedback is addressed. The one remaining gate is real-hardware testing before this ships in a tagged release.**
> 
> Re-reviewed the latest revision (d6b3c19) against the code, not just the replies:
> 
> - **Self-test** now covers v7_8 / v9 / v10 / **v11** — i.e. the live mainnet PoW. Call signatures match the function defs, `random_values` is reset per variant, and `salt` is reset between each HW and SW call (v10/v11 write back via `salt_pad`). With `xx=yy=2` the `salt_pad`/`r2` path is actually exercised, so the HW-vs-SW equivalence — including the `r2 = &c` (HW) vs `&b` (SW) source-buffer difference — is now guarded **fail-closed at startup**.
> - **`aes_expand_key`** borrows the context's `oaes_ctx` instead of doing a per-call `oaes_alloc`/`free`, with a NULL guard (defensive-only, since `cn_hash_context_create` returns NULL if the alloc fails). HW and SW paths are mutually exclusive per call, so borrowing is safe.
> - **x86 AES-NI detection** now reuses the project's tested `crypto::has_aesni()` CPUID path via a C wrapper — this also fixes the Cross-Mac x86_64 `___cpu_model` link error. The MinGW "multiple definition" of `SetLockPagesPrivilege` is fixed too.
> - **i686** no longer compiles the HW TU (`BUILD_64` gate); stale `#endif` guard comment updated.
> - The **x86 no-AES-NI behavior change** (graceful software fallback instead of hard-fail) is intentional and accepted — the fail-closed abort still fires on the case that matters (a HW path that produces a different hash than SW).
> 
> The NEON AES reformulation is byte-equivalent to x86 `aesenc`, the C/C++ linkage is correct, and the self-test is deterministic.
> 
> **Caveat — does not block merge, but blocks tagging a release:** the hardware path has not yet been exercised on real Apple Silicon or on AES-less ARMv8 (e.g. Raspberry Pi 4) hardware. CI only proves it compiles and that the `-march=armv8-a+crypto` toolchain probe passes. Before we cut a release that ships this, please capture at least one real arm64 run logging `cn_hardware_aes_supported() = 1` with a passing startup self-test, and ideally one AES-less ARMv8 run confirming the software fallback. The fail-closed self-test means any divergence would refuse to start rather than corrupt consensus — but we should still confirm the happy path on metal.
> 

Thanks for approval and the feedback :)
I can confirm that the built version is tested on my arm machines (one with ARM Cortex-A57 that has aes-hw, and one aes-less ARM Cortex-A72).
They both behave correctly and I observe a slight improvement on aes-capable one 😎👍

sn1f3rt:
Sounds good, Mac build issue is fixed on master, please rebase for CI to pass. PR should be good to ship with +1 from @ngeojiajun.

mmokhi:
> Sounds good, Mac build issue is fixed on master, please rebase for CI to pass. PR should be good to ship with +1 from @ngeojiajun. 

Done, just rebased it on master ✅
Thank you 👍

ngeojiajun (changes requested):
The PR itself is fine, but please address the following:
1. What is the point to extract out the code from slow-hash.c into slow-hash-impl.h and the c files? Where the C file just rename the symbol and import the HW/SW part.
In this case, it might be much easier to copy and paste the source in the c files. Moreover, putting non-inline source is considered a very bad idea since if the file is included in 2 files it will always introduce symbol conflict.
2. This PR have potential in causing performance regression on x86_64 platform due to the fact that all cn_* calls will require to go through branching before reaching a real function, and that will results on inefficient use of processor instruction cache.

Additional concerns (not blockers) @sn1f3rt :
1. Not sure if we use `--march=native` in our builds, but the changeset will defeat the purpose of using this flag which means generate and optimize the code onto the processor that it is compiling on.

mmokhi:
> The PR itself is fine, but please address the following:
> 
> 1. What is the point to extract out the code from slow-hash.c into slow-hash-impl.h and the c files? Where the C file just rename the symbol and import the HW/SW part.
> 

The main point was/is to have a cleaner implementation and mainly avoid code duplications :) 

> In this case, it might be much easier to copy and paste the source in the c files.

Indeed, easier but code duplications just adds maintainership costs from my experience. Of course it's only my one opinion and I'm open to change if you insist.

> Moreover, putting non-inline source is considered a very bad idea since if the file is included in 2 files it will always introduce symbol conflict.
> 

This is what I tried to avoid with #def guards, I could do #errors to make sure they'll not compile in advance if you prefer.

> 2. This PR have potential in causing performance regression on x86_64 platform due to the fact that all cn_* calls will require to go through branching before reaching a real function, and that will results on inefficient use of processor instruction cache.
> 
Been using the build artifacts on my machine (arm and x86, no visible regression that I witnessed). I think the theory you are referring to is the scale of 10^-6 (still true, but I argue negligible).
 
> 
> Additional concerns (not blockers) @sn1f3rt :
> 
> 1. Not sure if we use `--march=native` in our builds, but the changeset will defeat the purpose of using this flag which means generate and optimize the code onto the processor that it is compiling on.

sn1f3rt:
**1. Extracting the impl into `slow-hash-impl.h` and the `.c` shims / symbol conflict**

The "if the file is included in 2 files it will always introduce a symbol conflict" concern doesn't apply here. `slow-hash-impl.h` is included by both `slow-hash-hw.c` and `slow-hash-sw.c`, but each TU **renames the symbols before the include** (`cn_slow_hash_v11_hw` vs `cn_slow_hash_v11_sw`, gated by `CN_USE_SOFTWARE_AES`). That rename is precisely the mechanism that prevents the collision — it's deliberate, not a latent conflict.

The copy-paste alternative would duplicate several hundred lines of **consensus-critical** CryptoNight code across two files that then have to stay bit-identical forever. That's the more dangerous option for us. The reference implementation makes the same call in spirit — Monero keeps a *single* implementation and never maintains two copies; it just selects the path at runtime (samples below). So the single-source structure is intentional, and the symbol-rename approach additionally lets the HW and SW paths compile as separate translation units with different arch flags (HW with crypto extensions, SW without), which is cleaner than carrying crypto instructions through the software path.

**2. Potential x86_64 regression from branching / instruction cache**

The dispatch runs **once per full hash**, not per AES round. `CN_DISPATCH` expands to `if (cn_hardware_aes_supported()) {hw} else {sw}`, and `cn_hardware_aes_supported()` caches its result in a `static` — so steady-state cost is one cached load + one perfectly-predicted branch, after which we're inside `cn_slow_hash_v11_hw` for the entire 2 MB scratchpad fill and ~500K AES rounds. The hot loop carries no per-iteration branch.

This is the same runtime-dispatch design Monero has shipped for years, on both x86_64 and aarch64:

```c
// monero/src/crypto/slow-hash.c
static inline int force_software_aes(void) {
  static int use = -1;
  if (use != -1) return use;
  const char *env = getenv("MONERO_USE_SOFTWARE_AES");
  /* ... */
}

STATIC INLINE int check_aes_hw(void) {        // aarch64
#ifdef __APPLE__
    return 1;
#else
    static int supported = -1;
    if(supported < 0)
        supported = (getauxval(AT_HWCAP) & HWCAP_AES) != 0;
    return supported;
#endif
}
```

```c
// x86_64 cn_slow_hash
int useAes = !force_software_aes() && check_aes_hw();
/* ... */
// Two independent versions, one with AES, one without, to ensure that
// the useAes test is only performed once, not every iteration.
if(useAes) {
    for(i = 0; i < ITER / 2; i++) { pre_aes(); _c = _mm_aesenc_si128(_c, _a); post_aes(); }
} else {
    for(i = 0; i < ITER / 2; i++) { pre_aes(); aesb_single_round(/*...*/); post_aes(); }
}
```

```c
// aarch64 cn_slow_hash — same pattern, NEON path
int useAes = !force_software_aes() && check_aes_hw();
/* ... */
if(useAes) {
    for(i = 0; i < ITER / 2; i++) {
        pre_aes();
        _c = vaeseq_u8(_c, zero);
        _c = vaesmcq_u8(_c);
        _c = veorq_u8(_c, _a);
        post_aes();
    }
} else { /* aesb_single_round software path */ }
```

Monero's own comment — *"the useAes test is only performed once, not every iteration"* — is the whole point: the branch only costs anything if it's in the hot loop, and it isn't. This PR actually hoists it one level *higher* than Monero (Monero re-checks `useAes` ~3× per hash; here it's once at the wrapper and the `_hw` TU has zero `useAes` checks). The i-cache concern is therefore inverted — splitting HW/SW into separate functions keeps the hot path contiguous with no cold software code interleaved, which helps locality rather than hurting it. We can perform a before/after test if required.

**Additional: `-march=native`**

Our release/distributed builds don't use `-march=native` — the depends toolchain pins an explicit arch (`x86_64`, `armv8-a`) and adds `-maes` on x86 or `+crypto` on ARMv8, because release binaries have to stay portable across machines. `ARCH` only defaults to `native` for local from-source builds. So there's no conflict in the shipped binaries — and runtime dispatch is in fact what *lets* a portable (non-native) binary use hardware AES when the CPU has it, which is the whole goal here. For a local native build it's a negligible redundancy, not a regression.

ngeojiajun (approved):

PR #96, opened 2026-06-07
feat: --mining-threads-adaptive to re-autodetect when hashrate drops

mmokhi:
a laptop running nervad sees its hashrate drop after a heavy app is opened (browser, IDE, etc.). today the only way back to the local optimum is to restart nervad. with this flag, nervad notices the drop and re-tunes itself.

## Summary
- adds opt-in `--mining-threads-adaptive` flag (only honored with `--mining-threads 0`)
- after the existing autodetect locks in, samples the smoothed average from `m_last_hash_rates` every 60s. if it stays below 85% of baseline for 2 ticks, the existing autodetect re-runs and finds the new optimum. 30 min cooldown between re-runs to avoid flapping
- without the flag (default off), nothing changes

design notes:
- the adaptive check runs on the daemon's idle loop, not on a worker thread, and only when the feature is enabled. cost is one mutex acquire + sum of up to 19 `uint64_t` entries every 60s, all off the hot mining path
- re-trigger reuses the existing autodetect machinery (ramps 1 → N), simple and battle-tested
- reaction is asymmetric: we only re-detect on drops, not on improvements, to avoid flapping when conditions oscillate

## Test plan
- [ ] depends matrix is green
- [ ] start nervad with `--start-mining ADDR --mining-threads 0 --mining-threads-adaptive` on a multi-core machine, see the `Adaptive mining enabled` and `baseline locked at X H/s` log lines after the initial autodetect finishes
- [ ] push the host below 85% of baseline for at least ~2 minutes (heavy app, `stress-ng -c N`, etc.), watch the `sample N/2` log lines accumulate, then `hashrate dropped, re-running autodetect`, then the climb back up to the new optimum
- [ ] confirm no second re-trigger within 30 min (cooldown)
- [ ] run without the flag and confirm the existing autodetect path is unchanged

## Possible follow-ups (probably too far-fetched for this PR)
- **ramp down from N-1 instead of up from 1** on re-trigger. would cut the climb from ~50s to ~20-30s when contention has reduced the optimum, but needs modification of the upward-only `update_autodetection()`
- **hook into `merge_hr()` with an EWMA** instead of the 60s gate. faster reaction (~30-60s vs ~98s) but mixes adaptive logic into a single-purpose function
- **per-thread hashrate tracking** to identify which thread is contention-hit and only restart that one

### DIFF
--- a/src/cryptonote_basic/miner.cpp
+++ b/src/cryptonote_basic/miner.cpp
@@ -81,6 +81,10 @@
 #define AUTODETECT_WINDOW 10 // seconds
 #define AUTODETECT_GAIN_THRESHOLD 1.02f  // 2%
 
+#define ADAPTIVE_DROP_THRESHOLD 0.85f      // re-detect if H/s falls below 85% of baseline
+#define ADAPTIVE_DWELL_SAMPLES 2           // require this many consecutive low samples
+#define ADAPTIVE_COOLDOWN_MS (30 * 60 * 1000ull)  // wait 30 min between re-detections
+
 using namespace epee;
 
 #include "miner.h"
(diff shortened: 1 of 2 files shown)


mmokhi:
small update: pushed two drive-by refactor commits on top.

- `4b22230` **`refactor(miner): make m_threads_total atomic`** — it was `volatile uint32_t`, which doesn't actually provide cross-thread ordering by the C++ memory model. every other shared counter in `miner` (`m_template_no`, `m_threads_active`, `m_hashes`, `m_total_hashes`, `m_current_hash_rate`) is already `std::atomic`; this one was the odd one out. the adaptive code adds a new on_idle path reading/writing it, which made the gap a bit more relevant to close now.
- `a0f242f` **`refactor(miner): extract restart_workers() helper`** — the ~10-line "stop workers, wait, clear, reset, relaunch" block was duplicated between `update_autodetection()` and the new `update_adaptive_threads()`. now both call one helper. `start()` is intentionally left alone since it interleaves the launch with bg-mining setup, so its contract is genuinely different.

no behavior change in either; both have their own focused commit messages with full rationale.

PR #101, opened 2026-06-15
Feat: hard fork 13 (CryptoNight-Adaptive v6)

R0BC0D3R:
## Summary

Introduces Hard Fork 13 activating CryptoNight-Adaptive v6 (CNA v6) as
the new proof-of-work algorithm, along with a sync speed improvement and
transaction validation hardening.

### CryptoNight-Adaptive v6
- 4 MB scratchpad (up from 1 MB in v9–v12)
- Random VM program seeded per block for ASIC resistance
- Block data dependency via `get_cna_v6_data` for pool resistance

### Sliding window sync optimization
- `get_cna_v6_data` biases ~95% of block reads to the most recent
  100,000 blocks, a window small enough to reside in L3 cache
- Remaining ~5% draw from full chain history to preserve pool resistance
- Significantly reduces sync time on long chains post-HF13 regardless
  of total chain length

### Transaction validation hardening
- Additional validity checks on transaction inputs and outputs

## General test plan
- [x] Testnet HF13 activates at expected block height
- [x] CNA v6 PoW produces valid blocks and is accepted by peers
- [x] Sync from genesis completes without errors post-HF13
- [x] Existing wallets can send/receive after HF13 activates
- [x] Nodes running old software reject HF13 blocks (fork confirmed)

### Transaction validation rules (min-2-outputs, min-output-age)
- [ ] Post-HF13: a transaction with a single output is rejected; before the
      activation height a 1-output tx is still accepted (no early/retroactive enforcement)
- [ ] Post-HF13: a transaction spending an output younger than 10 blocks
      (CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE) is rejected
- [ ] Post-HF13: normal wallet send accepted, including the exact-change case
      (zero change → wallet adds a dummy 2nd output)
- [ ] Post-HF13: sweep_all / sweep to a single address accepted (≥2 outputs,
      decoys within the min-age bound) — this is the key check; the min-age
      boundary is exact, so normal sends/sweeps passing is the real validation
- [ ] Both rules enforce at block verification, not just mempool: an upgraded
      node rejects a *block* containing a 1-output or too-young-output tx
- [ ] Coinbase (single output) remains valid post-HF13


src/crypto/cna-vm.c
```diff
@@ -0,0 +1,179 @@
...
+    if (r == 0)
+        return x;
+    return (x >> r) | (x << (64 - r));
+}
+
+// High-quality 64-bit mixing function used by CN_OP_MIX.
+// Avalanche effect comparable to an AES round but portable and branchless.
+static uint64_t mix64(uint64_t val, uint32_t key_material)
+{
+    uint64_t k = (uint64_t)key_material * UINT64_C(0x9e3779b97f4a7c15);
```
ngeojiajun: these might be better marked as inline, but not sure

src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -2714,6 +2714,111 @@ void BlockchainLMDB::get_cna_v5_data(char *out, HC128_State *rng_state, uint64_t
   }
 }
 
+void BlockchainLMDB::get_cna_v6_data(char *out, HC128_State *rng_state, uint64_t height)
+{
+  assert(height > 0);
```
ngeojiajun: might be better use "CHECK_AND_ASSERT_MES" here, since assert will internally call  `abort()`  where will cause it to crash without message on certain platform

src/crypto/slow-hash.c
```diff
@@ -141,6 +144,12 @@ void cn_slow_hash_v11(cn_hash_context_t *ctx, const void *data, size_t length, c
                 cn_slow_hash_v11_sw(ctx, data, length, hash, iters, init_size_blk, xx, yy));
 }
 
+void cn_slow_hash_v13(cn_hash_context_t *ctx, const void *data, size_t length, char *hash, const uint8_t *seed)
+{
+    CN_DISPATCH(cn_slow_hash_v13_hw(ctx, data, length, hash, seed),
+                cn_slow_hash_v13_sw(ctx, data, length, hash, seed));
+}
+
```
ngeojiajun, line 153: In this case, the `cn_slow_hash_v13_hw` and `cn_slow_hash_v13_sw` is doing the exact same thing regardless of the it is using hardware or not. So it is better to just inline it here, instead of going through the dispatch

ngeojiajun (changes requested):

src/crypto/cna-vm.c
R0BC0D3R: Done. Both marked static inline in commit aef6cad.

src/blockchain_db/lmdb/db_lmdb.cpp
R0BC0D3R: Done. Replaced with CHECK_AND_ASSERT_MES(height > 0, , "get_cna_v6_data called with height == 0") in commit aef6cad.

src/crypto/slow-hash.c
R0BC0D3R: Good catch. v13 uses HC-128 for its VM, which is pure software, so hardware AES doesn't change the result. Added a self-test in cn_slow_hash_self_test (same commit) that runs both paths against identical inputs and asserts they match. This is now part of every startup check. Happy to also collapse the dispatch and call the SW path directly from cn_slow_hash_v13 if you'd prefer, but wanted to confirm the equivalence first.

ngeojiajun (approved):
The PR looks fine to me, just there is a potential risk of the pre-HF 13 transaction being mined into HF-13 block causing chain forks. Given the number of transactions in the chain, the risk should be minimal. 

It will be up to rest of the team to determine if we should address it or not.

src/cryptonote_core/blockchain.cpp
```diff
@@ -2726,6 +2743,11 @@ bool Blockchain::check_tx_outputs(const transaction& tx, tx_verification_context
         tvc.m_invalid_output = true;
         return false;
       }
+      if (!rct::isInMainSubgroup(rct::pk2rct(out_to_key.key)) || out_to_key.key == rct::rct2pk(rct::identity())) {
```
sn1f3rt: This check isn't gated by hardfork version, so it applies retroactively. `handle_block_to_main_chain` now calls `check_tx_outputs` on every non-coinbase tx of every block, with no checkpoint or quicksync skip (quicksync only bypasses the PoW longhash), so a from-genesis resync runs this against all ~4.3M blocks of history. Output keys were never subgroup-checked in consensus before, so one crafted historical tx with a torsion/identity output key that old nodes accepted would now halt resync and block new-node onboarding — testnet won't catch this (no adversarial history there). Please gate behind HF13, consistent with the min-2-outputs rule above:

```cpp
if (hf_version >= HF_VERSION_MIN_2_OUTPUTS) {
  if (!rct::isInMainSubgroup(rct::pk2rct(out_to_key.key)) || out_to_key.key == rct::rct2pk(rct::identity())) {
    MERROR_VER("Output public key is identity or not in main subgroup");
    tvc.m_invalid_output = true;
    return false;
  }
}
```

(Or land a documented full-chain scan proving zero violations. Gating is strictly safer: ungating gives no security benefit — you can't retroactively reject a long-buried tx, you only DoS honest resyncers — and it keeps a scalar-mult-by-`l` per output out of historical resync.)

src/cryptonote_core/blockchain.cpp
```diff
@@ -2914,13 +2941,27 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc,
     // make sure tx output has key offset(s) (is signed to be used)
     CHECK_AND_ASSERT_MES(in_to_key.key_offsets.size(), false, "empty in_to_key.key_offsets in transaction with id " << get_transaction_hash(tx));
 
+    for (size_t k = 1; k < in_to_key.key_offsets.size(); ++k) {
```
sn1f3rt: Same retroactivity concern as the output-key check — `check_tx_inputs` runs on all historical blocks during resync. Master never rejected duplicate ring members, so any historical tx with `key_offsets[k] == 0` that old consensus accepted would now fail resync and split off new nodes. `hf_version` is already in scope here; please gate behind HF13:

```cpp
if (hf_version >= HF_VERSION_MIN_2_OUTPUTS) {
  for (size_t k = 1; k < in_to_key.key_offsets.size(); ++k) {
    if (in_to_key.key_offsets[k] == 0) {
      MERROR_VER("Transaction input has duplicate ring member");
      tvc.m_verifivation_failed = true;
      return false;
    }
  }
}
```

src/cryptonote_core/blockchain.cpp
```diff
@@ -2914,13 +2941,27 @@ bool Blockchain::check_tx_inputs(transaction& tx, tx_verification_context &tvc,
...
+    }
+
     if(have_tx_keyimg_as_spent(in_to_key.k_image))
     {
       MERROR_VER("Key image already spent in blockchain: " << epee::string_tools::pod_to_hex(in_to_key.k_image));
       tvc.m_double_spend = true;
       return false;
     }
 
+    if (!rct::isInMainSubgroup(rct::ki2rct(in_to_key.k_image)) || rct::ki2rct(in_to_key.k_image) == rct::identity()) {
```
sn1f3rt: Same fix needed. Honest key images are already in the main subgroup (`hash_to_ec` clears the cofactor via `ge_mul8`), but torsion key images were historically acceptable to CryptoNote/Monero consensus, so if any exist in Nerva's early history this halts every from-genesis resync. Gate behind HF13:

```cpp
if (hf_version >= HF_VERSION_MIN_2_OUTPUTS) {
  if (!rct::isInMainSubgroup(rct::ki2rct(in_to_key.k_image)) || rct::ki2rct(in_to_key.k_image) == rct::identity()) {
    MERROR_VER("Key image is identity or not in main subgroup");
    tvc.m_verifivation_failed = true;
    return false;
  }
}
```

(If you'd rather not overload `HF_VERSION_MIN_2_OUTPUTS` for all three, add a dedicated `HF_VERSION_TX_KEY_VALIDATION 13` in `cryptonote_config.h` — clearer intent, same value.)

src/crypto/slow-hash.c
```diff
@@ -194,6 +203,11 @@ cn_hash_context_t *cn_hash_context_create(void)
         cn_hash_context_free(ctx);
         return NULL;
     }
+    ctx->cna_scratchpad_is_mapped = allocate_hugepage(CN_SCRATCHPAD_MEMORY_V13, (void **)&(ctx->cna_scratchpad));
```
sn1f3rt, line 343: The context is `malloc`'d un-zeroed at line 192. If this new `allocate_hugepage` for `cna_scratchpad` (or the `scratchpad` alloc above it) fails, the `cn_hash_context_free(ctx)` cleanup reads the still-uninitialized `salt`/`cna_scratchpad` pointers and can `free_hugepage` a garbage pointer. Pre-existing pattern, but this PR adds another un-zeroed pointer to it. Simplest fix is to zero-initialize at allocation:

```cpp
cn_hash_context_t *ctx = (cn_hash_context_t *)calloc(1, sizeof(cn_hash_context_t));
```

sn1f3rt (commented):
Reviewed the HF13 changes in depth against their callers and bounds. The PoW/VM (`cna-vm`, `cn_slow_hash_v13`), the sliding-window `get_cna_v6_data`, the v13 longhash dispatch, and the gated min-2-outputs / min-output-age rules are correct and deterministic. Coinbase is correctly excluded from `check_tx_outputs` (the loop is over `bl.tx_hashes` only), and the "forbid non-zero amounts" rule is safe retroactively because master's `check_tx_inputs` already rejects `RCTTypeNull` non-coinbase txs, so the chain is provably all-RingCT.

**One blocking item, raised inline:** three new validation checks (output-key subgroup, key-image subgroup, duplicate ring member) are **not** gated by hardfork version. Because `handle_block_to_main_chain` runs `check_tx_outputs`/`check_tx_inputs` on every block with no checkpoint or quicksync skip (quicksync only bypasses the PoW longhash), they apply retroactively to the whole chain on any from-genesis resync. A single crafted historical tx that old consensus accepted would halt resync and prevent new nodes from onboarding — and testnet won't surface it. Please gate all three behind HF13 (or land a documented full-chain scan proving zero violations). One un-zeroed-context cleanup nit in `slow-hash.c` as well.

### Suggested: add a Known Answer Test (KAT) for v13

The v13 self-test only checks **HW path == SW path on the same machine**. For a consensus PoW that has two blind spots:

- If both paths share the same bug, `hw == sw` still passes — the test can't see it.
- It doesn't pin the *absolute* output, so a future refactor, compiler change, or endianness difference that silently alters the hash (and forks the network) won't be caught — both paths just change together.

A KAT closes both by asserting a fixed input hashes to an exact, hardcoded value. The self-test already sets up a deterministic input (`seed = {0}`, zeroed `salt` and `random_values`), so it's a two-line addition:

```c
static const uint8_t v13_kat[HASH_SIZE] = { /* capture from a known-good build */ };
if (memcmp(hw, v13_kat, HASH_SIZE) != 0) ok = 0;
```

The one wrinkle is chicken-and-egg: generate the expected bytes from a reference build, then hardcode them. Best practice is to generate and cross-check on at least two architectures (e.g. x86_64 and aarch64) before locking the constant — that simultaneously proves the cross-platform determinism a PoW hardfork most needs, which the HW/SW self-test alone doesn't guarantee.

src/cryptonote_core/blockchain.cpp
R0BC0D3R: Good catch, agreed. Gated behind a new HF_VERSION_TX_KEY_VALIDATION (=13) in 334accb so it only applies from HF13 onward and historical resync is unchanged. Used a dedicated constant rather than overloading MIN_2_OUTPUTS for clearer intent, as you suggested.

src/cryptonote_core/blockchain.cpp
R0BC0D3R: Done in 334accb. Gated behind HF_VERSION_TX_KEY_VALIDATION (=13), same as the output-key check.

src/cryptonote_core/blockchain.cpp
R0BC0D3R: Done in 334accb. Also gated behind HF_VERSION_TX_KEY_VALIDATION (=13). All three key/ring checks now share the one gate.

src/crypto/slow-hash.c
R0BC0D3R: Fixed in 334accb. Context is now calloc'd so the pointer fields start NULL and a failed allocation can't leave cn_hash_context_free freeing a garbage pointer.

sn1f3rt (approved):
Approving — good to merge subject to testnet testing and the KAT being added.

PR #102, opened 2026-06-17
fix(common): defer Ctrl-C handling off signal context (FreeBSD coredump)

mmokhi:
On FreeBSD, pressing Ctrl-C on a running daemon aborts in libthr instead of stopping cleanly:

```
Fatal error 'thread 0x... was already on queue.'
  at line 285 in file /usr/src/lib/libthr/thread/thr_cond.c (errno = 19)
Abort trap (core dumped)
```

`signal_handler::posix_handler` does the real work in signal context: `handle_signal()` takes a `boost::mutex` and calls `m_handler` (`stop_p2p` -> `send_stop_signal` for the daemon). That touches pthread mutex/condvar from the handler, which is not async-signal-safe; libthr aborts there, glibc happens not to. At an idle interactive prompt the main loop sits in the console reader's condvar wait (`console_handler.h`), which is the thread SIGINT interrupts.

It's the same `send_stop_signal()` the `exit` command reaches (`stop_daemon` -> `on_stop_daemon`), so Ctrl-C and `exit` are meant to behave the same.

## Fix

Move the work out of signal context. `posix_handler` now only writes the signal number to a self-pipe (`write()` is async-signal-safe); a worker thread, started once via `boost::call_once`, drains the pipe and runs `handle_signal()`/`m_handler` in normal thread context. The interrupted thread no longer makes any pthread call from the handler.

Self-pipe rather than `sem_init` (unnamed semaphores are unsupported on macOS); `pipe()`/`read()`/`write()` work on Linux, FreeBSD and macOS. Windows path unchanged.

### DIFF
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -183,13 +183,9 @@ namespace tools
       }
       return r;
 #else
-      // The real handler must never run in async-signal context: it locks a
-      // mutex and dispatches into arbitrary C++ (logging, allocation, more
-      // locks). That is undefined behaviour per POSIX. glibc tolerates it in
-      // practice, but FreeBSD's libthr aborts (coredump) when pthread state is
-      // touched from a signal handler. So the signal handler only does
-      // async-signal-safe work (a single write() to a self-pipe) and a
-      // dedicated thread runs m_handler in normal thread context.
+      // Mutex locks + arbitrary C++ in m_handler aren't async-signal-safe;
+      // FreeBSD's libthr aborts when pthread state is touched from a handler.
+      // Defer the real work to a thread fed by an async-signal-safe write().
       boost::call_once(m_dispatch_once, &signal_handler::start_dispatch_thread);
       m_handler = t;
 
@@ -222,21 +218,15 @@ namespace tools
       return TRUE;
     }
 #else
-    /*! \brief async-signal-safe handler for NIX
-     *
-     * Runs in signal context, so it must only call async-signal-safe
-     * functions. It records the signal number on a self-pipe; the dispatch
-     * thread (signal_dispatch_thread) picks it up and runs the real handler.
-     */
+    /*! \brief async-signal-safe NIX handler: records the signal on the
+     *  self-pipe; signal_dispatch_thread runs the real handler. */
     static void posix_handler(int type)
     {
       const unsigned char sig = (unsigned char)type;
-      // write() is async-signal-safe; ignore the result (errno is preserved by
-      // the kernel only across this call, and there is nothing safe to do on
-      // failure inside a signal handler anyway).
+      // write() is async-signal-safe; nothing safe to do on failure.
       const int saved_errno = errno;
       while (write(m_pipe[1], &sig, 1) < 0 && errno == EINTR) {}
-      errno = saved_errno;
+      errno = saved_errno; // restore for the interrupted code
     }
 
     /*! \brief sets up the self-pipe and the dispatch thread (once) */
--- a/src/common/util.h
+++ b/src/common/util.h
@@ -33,10 +33,13 @@
 
 #include <boost/thread/locks.hpp>
 #include <boost/thread/mutex.hpp>
+#include <boost/thread/thread.hpp>
+#include <boost/thread/once.hpp>
 #include <boost/optional.hpp>
 #include <system_error>
 #include <csignal>
 #include <cstdio>
+#include <cstring>
 #include <functional>
 #include <memory>
 #include <string>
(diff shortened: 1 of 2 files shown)


mmokhi:
Tested on FreeBSD 14.2: built the daemon and ran it interactively, then sent SIGINT at the prompt (the original repro). It now shuts down cleanly with exit code 0 — no `was already on queue` libthr abort and no core dump. Before this change the same path gave `Abort trap (core dumped)`.

src/common/util.h
```diff
@@ -176,6 +183,16 @@ namespace tools
       }
       return r;
 #else
+      // The real handler must never run in async-signal context: it locks a
+      // mutex and dispatches into arbitrary C++ (logging, allocation, more
+      // locks). That is undefined behaviour per POSIX. glibc tolerates it in
+      // practice, but FreeBSD's libthr aborts (coredump) when pthread state is
+      // touched from a signal handler. So the signal handler only does
+      // async-signal-safe work (a single write() to a self-pipe) and a
+      // dedicated thread runs m_handler in normal thread context.
```
sn1f3rt: nit: this 7-line block is also stated in the commit message and the `posix_handler` brief below — 3 lines carry the essential "why":

```suggestion
      // Mutex locks + arbitrary C++ in m_handler aren't async-signal-safe;
      // FreeBSD's libthr aborts when pthread state is touched from a handler.
      // Defer the real work to a thread fed by an async-signal-safe write().
```

src/common/util.h
```diff
@@ -206,10 +222,49 @@ namespace tools
       return TRUE;
     }
 #else
-    /*! \brief handler for NIX */
+    /*! \brief async-signal-safe handler for NIX
+     *
+     * Runs in signal context, so it must only call async-signal-safe
+     * functions. It records the signal number on a self-pipe; the dispatch
+     * thread (signal_dispatch_thread) picks it up and runs the real handler.
+     */
```
sn1f3rt: nit: the body restates the block in `install()`; a 2-line brief is enough:

```suggestion
    /*! \brief async-signal-safe NIX handler: records the signal on the
     *  self-pipe; signal_dispatch_thread runs the real handler. */
```

src/common/util.h
```diff
@@ -206,10 +222,49 @@ namespace tools
...
     static void posix_handler(int type)
     {
-      handle_signal(type);
+      const unsigned char sig = (unsigned char)type;
+      // write() is async-signal-safe; ignore the result (errno is preserved by
+      // the kernel only across this call, and there is nothing safe to do on
+      // failure inside a signal handler anyway).
+      const int saved_errno = errno;
+      while (write(m_pipe[1], &sig, 1) < 0 && errno == EINTR) {}
+      errno = saved_errno;
```
sn1f3rt: nit: "errno is preserved by the kernel only across this call" is a bit muddled — the point of the save/restore is just to not clobber errno for the interrupted code. One line + an inline note reads cleaner:

```suggestion
      // write() is async-signal-safe; nothing safe to do on failure.
      const int saved_errno = errno;
      while (write(m_pipe[1], &sig, 1) < 0 && errno == EINTR) {}
      errno = saved_errno; // restore for the interrupted code
```

sn1f3rt (approved):
Left a few nits regarding trimming comments, rest LGTM 👍

src/common/util.h
sn1f3rt: nit: this 7-line block is also stated in the commit message and the `posix_handler` brief below — 3 lines carry the essential "why":

```suggestion
      // Mutex locks + arbitrary C++ in m_handler aren't async-signal-safe;
      // FreeBSD's libthr aborts when pthread state is touched from a handler.
      // Defer the real work to a thread fed by an async-signal-safe write().
```

src/common/util.h
sn1f3rt: nit: the body restates the block in `install()`; a 2-line brief is enough:

```suggestion
    /*! \brief async-signal-safe NIX handler: records the signal on the
     *  self-pipe; signal_dispatch_thread runs the real handler. */
```

src/common/util.h
sn1f3rt: nit: "errno is preserved by the kernel only across this call" is a bit muddled — the point of the save/restore is just to not clobber errno for the interrupted code. One line + an inline note reads cleaner:

```suggestion
      // write() is async-signal-safe; nothing safe to do on failure.
      const int saved_errno = errno;
      while (write(m_pipe[1], &sig, 1) < 0 && errno == EINTR) {}
      errno = saved_errno; // restore for the interrupted code
```

sn1f3rt (approved):
Left a few nits regarding trimming comments, rest LGTM

mmokhi:
> Left a few nits regarding trimming comments, rest LGTM

Thanks for the suggestions :blush: 
I addressed them :heavy_check_mark:

old-greywolf:
i tested the FreeBSD build from this PR on FreeBSD 15.0-RELEASE-p5 and Ctrl-C now shuts down cleanly, no more libthr abort.

PR #104, opened 2026-06-19
perf: optimize Keccak permutation (port Monero #8262)

R0BC0D3R:
Port of Monero #8262: replace the table-driven keccakf loop with the unrolled
permutation (hardcoded rho-pi offsets + THETA/CHI macros) and drop the
keccakf_rotc/keccakf_piln lookup tables.

Output is bit-identical (same Keccak-f[1600]), so block hashes and CNA PoW are
unaffected; cn_fast_hash is faster. Speeds block/tx verification and PoW
pre-processing.

**Testing:** synced from block 0 past HF7 (CNA v2); ran with mining enabled, 
validated 20+ incoming newly-mined blocks and completed a wallet-to-wallet
transfer. All clean, no verify errors.

### DIFF
--- a/src/crypto/keccak.c
+++ b/src/crypto/keccak.c
@@ -31,53 +31,82 @@ const uint64_t keccakf_rndc[24] =
     0x8000000000008080, 0x0000000080000001, 0x8000000080008008
 };
 
-const int keccakf_rotc[24] = 
-{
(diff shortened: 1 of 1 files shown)


ngeojiajun (approved):
Seems sensible to me

sn1f3rt (approved):

PR #105, opened 2026-06-21
HF13: make CNA v6 memory-bound (closer to 1 CPU = 1 vote)

mmokhi:
## What

Reworks the v6 VM so the hash is gated by memory, not raw compute. The old VM was mostly register math (~78% ALU), so the fastest, widest cores just pulled ahead. Memory latency runs at about the same speed on a small computer, a laptop or a big grid, so leaning on it brings per-core hashrate closer together. Aim is to tighten 1 CPU = 1 vote without touching pool or ASIC resistance.

## Changes (all in the VM)

- **Opcode mix.** Now a memory-op majority ([51,63]% scratchpad reads/writes), with the ratio drawn from the per-block seed so the shape shifts every block and no ASIC can bake in a fixed one. ALU ops stay evenly weighted, so IMUL/MIX keep their diffusion.
- **Pointer chase.** Each scratchpad access folds its result into the next address, so the reads form a chain the CPU can't run ahead of. Same idea as CryptoNight's `state_index`. This is what evens out per-core throughput.
- **Scratchpad 4 MB -> 8 MB.** Under full multi-core mining, 8 MB per thread overflows the L3-per-core budget on nearly every machine, so they all fall back to main-memory latency. See the compromise below.

## Why 8 MB and not bigger

8 MB is a deliberate middle. Bigger (16/32 MB) would also catch huge-cache desktop chips, but it pushes up per-block verify and IBD time, which the sliding-window work was added to protect. 8 MB is a 2x step in the spirit of the earlier 1->4 MB bump, and it keeps that sync work intact.

## Safety

- The random per-block program is unchanged, so ASIC resistance stands.
- Program-generation draw order, the read/write split and the chain logic are deterministic; HW and SW AES paths share `cn_vm_*`, so they stay in lockstep (self-test still passes HW == SW).

## Open items

- [ ] **Activation heights (merge gate).** Testnet activates HF13 at 2100, stagenet at 800, mainnet at 4,300,000. Need to confirm testnet/stagenet haven't already crossed those on the 4 MB v6, otherwise this is a mid-fork PoW rewrite and they split. Operational check, not a code fix.
- [ ] **Known-answer vector (before freeze).** Self-test only checks HW == SW, not the hash value. Add a pinned `(input, seed) -> hash` once the algorithm is final, as a drift guard. Separate commit.

### DIFF
--- a/src/crypto/cna-vm.c
+++ b/src/crypto/cna-vm.c
@@ -42,6 +42,12 @@ void cn_vm_generate_program(cn_vm_program_t *prog, const uint8_t seed[32])
     HC128_State rng;
     // Use seed as both key and IV — first 16 bytes key, last 16 bytes IV.
     HC128_Init(&rng, (unsigned char *)seed, (unsigned char *)(seed + 16));
+    // HC128_Init sets up P/Q but does NOT fill the keystream buffer; the first
+    // HC128_NextKeys generates it. Without this, the first 16 HC128_U32 reads
+    // come from uninitialised stack (keystream[0..15]), so the program is
+    // non-deterministic and the HW and SW paths disagree. Every other HC128
+    // caller (get_cna_v5_data/get_cna_v6_data) does Init then NextKeys first.
+    HC128_NextKeys(&rng);
 
     size_t key_idx = 0;
 
--- a/src/crypto/cna-vm.c
+++ b/src/crypto/cna-vm.c
@@ -45,11 +45,29 @@ void cn_vm_generate_program(cn_vm_program_t *prog, const uint8_t seed[32])
 
     size_t key_idx = 0;
 
+    // Lean the mix toward scratchpad ops. A memory-bound program hashes at the
+    // same per-core rate on a small computer, a laptop or a big grid, which is
+    // what 1 CPU = 1 vote needs. Two rules keep it honest:
+    //   - mem_pct comes from the seed, so the ratio shifts every block and no
+    //     ASIC can bake in a fixed one.
+    //   - ALU ops stay evenly weighted, so IMUL/MIX keep their bite. Multiply is
+    //     our best ASIC repellent, so we don't water it down.
+    static const uint8_t alu_ops[7] = {
+        CN_OP_IADD_RS, CN_OP_ISUB, CN_OP_IMUL, CN_OP_IXOR,
+        CN_OP_IROR,    CN_OP_CBRANCH, CN_OP_MIX
+    };
+    // Memory-op share for this program: [48, 63]%. Always a majority, never fixed.
+    const uint32_t mem_pct = 48U + HC128_U32(&rng, &key_idx, 16U);
+
     for (int i = 0; i < CN_PROGRAM_SIZE; i++)
     {
         cn_vm_instruction_t *ins = &prog->instructions[i];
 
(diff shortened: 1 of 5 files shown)


old-greywolf:
the nervad (daemon) in the nerva-macos-armv8 Build Artifact returns:  Hardware AES self-test FAILED: the AES intrinsic path produced a different hash than the software path. Refusing to start with a broken hashing implementation. Re-run with NERVA_FORCE_SOFTWARE_AES=1 to bypass and use software AES only.

mmokhi:
> the nervad (daemon) in the nerva-macos-armv8 Build Artifact returns:  Hardware AES self-test FAILED: the AES intrinsic path produced a different hash than the software path. Refusing to start with a broken hashing implementation. Re-run with NERVA_FORCE_SOFTWARE_AES=1 to bypass and use software AES only.

Thanks for testing 😊
Technically I'm not changing any part that impacts hw or sw aes in my PR, I wonder if we're hitting a deeper part here in your test.

I'm gonna look deeper to see from a vanilla hf13 comparatively with my branch ✅

mmokhi:
@old-greywolf, you were right, this is a real bug, and I found the root cause.

In `cn_vm_generate_program` (the v13 VM) we seed the per-block program with HC128, but we were reading from it straight after `HC128_Init` without calling `HC128_NextKeys` first. `Init` sets up the cipher state but doesn't fill the keystream buffer, the first `NextKeys` does. So the first 16 RNG words were uninitialised stack, the program came out non-deterministic, and the HW (AES-NI / NEON) and SW paths hashed different things. That's why the self-test failed on some machines/runs and not others.

Every other HC128 caller in the tree already does Init then NextKeys before reading; v13 was the odd one out. One-line fix (`d769d07`): call `HC128_NextKeys` right after Init. Self-test now passes deterministically, HW == SW, confirmed on a fresh CI-built `nerva-linux-x86_64` on an AES-NI box.

Feel free to test it again please.

sn1f3rt (approved):
Code looks correct and deterministic — verified the HC128 keystream fix, the seed-driven opcode mix, and the pointer chase (bounds, 32-bit address math, HW/SW lockstep). Approving.

Mergeable subject to a macos-armv8 (NEON) re-test confirming HW == SW on the platform that originally failed. No testnet/stagenet split risk since HF13 hasn't been activated there yet. KAT will be added to the main HF13 PR (#101).

old-greywolf:
> @old-greywolf, you were right, this is a real bug, and I found the root cause.
> 
> In `cn_vm_generate_program` (the v13 VM) we seed the per-block program with HC128, but we were reading from it straight after `HC128_Init` without calling `HC128_NextKeys` first. `Init` sets up the cipher state but doesn't fill the keystream buffer, the first `NextKeys` does. So the first 16 RNG words were uninitialised stack, the program came out non-deterministic, and the HW (AES-NI / NEON) and SW paths hashed different things. That's why the self-test failed on some machines/runs and not others.
> 
> Every other HC128 caller in the tree already does Init then NextKeys before reading; v13 was the odd one out. One-line fix (`d769d07`): call `HC128_NextKeys` right after Init. Self-test now passes deterministically, HW == SW, confirmed on a fresh CI-built `nerva-linux-x86_64` on an AES-NI box.
> 
> Feel free to test it again please.

fixed.  good work.

Issue #106, opened 2026-06-23
iOS: ship a multi-slice .xcframework (device + simulator)


sn1f3rt:
Follow-up to #99, which added iOS (arm64) build support and ships a device-only, merged `libnerva_wallet.a` + `wallet/api` headers as a pre-release artifact.

The proper distribution format for an iOS static library is an `.xcframework` bundling separate slices per platform, so consumers can link the same package whether they build for a physical device or the simulator. Right now we only produce the device slice (`arm64-apple-ios`); the simulator slice (`arm64-apple-ios-simulator`) is missing, so the artifact can't be used in a simulator build.

### Scope
- Add a **simulator depends host** for `arm64-apple-ios-simulator` (SDK `iphonesimulator`, e.g. `-target arm64-apple-ios-simulator` / `--sdk iphonesimulator`), alongside the existing `aarch64-apple-ios` device host.
- Build `wallet_api` for both slices and produce a merged static lib per slice (same `libtool -static` approach as #99, including the boost test-framework exclusion).
- Assemble the final bundle with `xcodebuild -create-xcframework -library <device>.a -headers <inc> -library <sim>.a -headers <inc> -output libnerva_wallet.xcframework`.
- Upload the `.xcframework` as the iOS artifact and drop the device-only / pre-release caveat from `README-ios.txt`.
- Add the new simulator target name to the `targets` list in `post-artifact-links.yml` (coupled to the depends matrix), as done for `nerva-ios-armv8` in #99.

### Notes / gotchas
- The depends **host-triple parsing** is fiddly for the `-simulator` suffix: device and simulator are both `arm64` and differ only by platform, so the host name/triple split in `contrib/depends` needs care to not collapse them.
- `lipo` **cannot** merge same-arch/different-platform slices (arm64 device vs arm64 simulator) - that's exactly why `xcframework` is required rather than a fat `.a`.

### Acceptance criteria
- [ ] CI produces `libnerva_wallet.xcframework` with both `ios-arm64` and `ios-arm64-simulator` slices.
- [ ] Links cleanly in a minimal sample app for both a device and a simulator target.

sn1f3rt:
Done in #107

PR #107, opened 2026-06-24
feat(ios): ship a device + simulator .xcframework

mmokhi:
Implements #106. Follow-up to #99, which shipped a **device-only** merged `libnerva_wallet.a`. iOS consumers also need an **arm64 simulator** slice, and `lipo` can't combine same-arch/different-platform slices — so the correct distribution format is an `.xcframework`.

## What changed

**depends — add the simulator host (`aarch64-apple-iossimulator`)**
- Two `config.sub` layers: Nerva's own `config.sub` accepts the dash-less `iossimulator` triple (the dashed `-simulator` form is rejected), so `canonical_host` parses; `Makefile` maps both `ios` and `iossimulator` → `hosts/ios.mk`.
- Packages' *bundled* `config.sub` are too old to know `iossimulator`, so autotools `configure` gets the plain `aarch64-apple-ios` `--host` (which they accept, same as device); the simulator target is carried purely in the compiler flags. `ios.mk` switches SDK + triple (`iphonesimulator` + `arm64-apple-ios<min>-simulator`); `toolchain.cmake.in` picks `CMAKE_OSX_SYSROOT` off `@HOST@`. This mirrors how monero_c keeps the platform in CC/CMake, not the autotools host triple.

**CI (`depends.yml`) — new `build-ios` job**
- Builds `wallet_api` for both hosts (`aarch64-apple-ios` device, `aarch64-apple-iossimulator` simulator), and merges each into one linkable per-slice static lib via the #99 `libtool -static` recipe (minus boost's test-only frameworks, which are unused and collide on member names). Each slice is sanity-checked to actually carry the wallet API symbols (`nm … | grep WalletManagerFactory`).
- Headers are copied into a **headers-only staging dir** before `-headers`, so the framework ships just the `.h` and not the wallet `.cpp` sources.
- `xcodebuild -create-xcframework` bundles both slices into `libnerva_wallet.xcframework`, then the *Verify both slices present* step asserts the `Info.plist` carries both `ios-arm64` and `ios-arm64-simulator`.
- Uploaded as the `nerva-ios-armv8` artifact.

## Acceptance criteria
- [x] CI produces `libnerva_wallet.xcframework` with both `ios-arm64` and `ios-arm64-simulator` slices (asserted in *Verify both slices present*; `xcodebuild -create-xcframework` itself only succeeds on valid, correctly-typed slices).
- [x] **Linkability, verified structurally in CI:** each slice is a valid platform Mach-O (enforced by the xcframework assembly) and carries the wallet API symbols (`nm … | grep WalletManagerFactory`). A literal "build a sample Xcode app" link is left as a one-time downstream/reviewer check (it needs Xcode and a real app target, which a CLI link can't faithfully stand in for) — the natural place for that is the cake_wallet / monero_c (#188) integration.

### DIFF
--- a/.github/workflows/depends.yml
+++ b/.github/workflows/depends.yml
@@ -150,21 +150,19 @@ jobs:
                     path: |
                         build/${{ matrix.toolchain.host }}/release/bin/nerva*
 
-    # iOS has no Linux cross path (no iphoneos SDK to host, no Linux->iphoneos
-    # toolchain), so it builds native on macOS with the runner's Xcode. Deps still
-    # go through contrib/depends (apple-ios host). iOS ships a wallet library, not
-    # daemons, so we build the wallet_api target.
+    # iOS has no Linux cross path (no iphoneos SDK to host), so it builds native on
+    # macOS with the runner's Xcode. Deps go through contrib/depends; iOS ships a
+    # wallet library (wallet_api), not daemons. We build two slices - arm64 device
+    # and arm64 simulator - and bundle them into one .xcframework, since lipo can't
+    # combine same-arch/different-platform slices.
     build-ios:
+        name: ARMv8 iOS (xcframework)
         runs-on: macos-14
-        timeout-minutes: 120
-        strategy:
-            fail-fast: false
-            matrix:
-                toolchain:
-                    -   name: "ARMv8 iOS"
-                        host: "aarch64-apple-ios"
-                        artifact-name: "nerva-ios-armv8"
-        name: ${{ matrix.toolchain.name }}
+        timeout-minutes: 180
+        env:
+            DEVICE_HOST: aarch64-apple-ios
+            SIM_HOST: aarch64-apple-iossimulator
+            ARTIFACT: nerva-ios-armv8
         steps:
             -   name: Install dependencies
                 run: brew install autoconf automake libtool pkg-config gperf ccache coreutils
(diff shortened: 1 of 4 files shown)


old-greywolf:
the nerva-windows-x32 daemon, when launched on a 32-bit Windows physical computer (not a virtual machine) with an Intel Atom Processor Z3740 that includes the Intel AES New Instructions (Intel AES-NI), reports:  Hardware AES (AES-NI / ARMv8 Crypto Extensions) was not detected. Falling back to the software AES path; hashing will be substantially slower.

mmokhi:
@old-greywolf the `nervad.exe` crash you hit isn't related to this PR — it's a connection-teardown race in epee (with a sibling block-cache race that turned up on a Linux node). Both are fixed in #108, so this iOS PR can go ahead independently of it.

(The 32-bit "Hardware AES … not detected" line is a separate, known thing — the i686 build just doesn't compile the AES-NI path — handled on its own, not here.)

.github/workflows/depends.yml
```diff
@@ -184,46 +182,53 @@ jobs:
...
+                        nm "build/$H/libnerva_wallet.a" 2>/dev/null | grep -q WalletManagerFactory \
+                          || { echo "no wallet_api symbols in $H slice"; exit 1; }
+                    done
+            -   name: Assemble xcframework
+                run: |
+                    STAGE="$GITHUB_WORKSPACE/$ARTIFACT"
+                    mkdir -p "$STAGE"
+                    xcodebuild -create-xcframework \
+                        -library "build/$DEVICE_HOST/libnerva_wallet.a" -headers src/wallet/api \
+                        -library "build/$SIM_HOST/libnerva_wallet.a"    -headers src/wallet/api \
```
sn1f3rt: `-headers src/wallet/api` copies the *entire* directory into each slice's `Headers/`, not just the headers. `src/wallet/api` also holds 9 `.cpp` files (and a `.txt`), so the distributed xcframework ends up shipping the implementation sources. The prior recipe was deliberately headers-only (`cp src/wallet/api/*.h`). Consider staging the `.h` files into a headers-only dir first and pointing `-headers` at that to avoid leaking source into the framework.

.github/workflows/depends.yml
```diff
@@ -184,46 +182,53 @@ jobs:
...
+                    echo "$INFO"
+                    echo "$INFO" | grep -q '"ios-arm64"'           || { echo "missing device slice";    exit 1; }
+                    echo "$INFO" | grep -q '"ios-arm64-simulator"' || { echo "missing simulator slice"; exit 1; }
             -   uses: actions/upload-artifact@v7
                 if: github.event_name != 'schedule'
                 with:
-                    name: ${{ matrix.toolchain.artifact-name }}
-                    path: ${{ matrix.toolchain.artifact-name }}/
+                    name: ${{ env.ARTIFACT }}
+                    path: nerva-ios-armv8/
```
sn1f3rt: This `path:` is hardcoded while `name:` just above uses `${{ env.ARTIFACT }}`. They agree today, but if `ARTIFACT` is ever changed this upload silently breaks (path not found). Suggest `path: ${{ env.ARTIFACT }}/` for consistency with `STAGE` and the verify step, which both use `$ARTIFACT`.

sn1f3rt (commented):
Clean, well-factored change. I traced the build mechanism end-to-end and it holds up: config.sub preserves the `iossimulator` triple so `full_host_os` drives the simulator SDK in `hosts/ios.mk`, `@HOST@` is substituted so the toolchain picks `iphonesimulator`, the Makefile `--host` override (`aarch64_ios_host=aarch64-apple-ios`) does reach `./configure` via `$($(1)_type)_host` while the prefix stays distinct, and the `"ios-arm64"` vs `"ios-arm64-simulator"` slice check is not a false-positive (the trailing quote disambiguates). No correctness issues.

One note on the PR description, not the code: it says this ships "the `nerva-ios` artifact (renamed from `nerva-ios-armv8`, now universal) and update[s] the `post-artifact-links` target." Neither actually happened - `ARTIFACT` is still `nerva-ios-armv8` and `post-artifact-links.yml` is untouched (its existing entry still resolves, so nothing is broken). The commit message correctly says the post-artifact-links entry is unchanged, so it's just the PR body that's stale. Worth reconciling so reviewers aren't hunting for a rename that isn't there.

Two minor inline nits below.

.github/workflows/depends.yml
mmokhi: both fixed: headers now staged into a headers-only dir before -headers so the framework ships just the .h and not the wallet api .cpp sources, and the upload path uses $ARTIFACT to match name: and STAGE.

.github/workflows/depends.yml
mmokhi: both fixed: headers now staged into a headers-only dir before -headers so the framework ships just the .h and not the wallet api .cpp sources, and the upload path uses $ARTIFACT to match name: and STAGE.

sn1f3rt (approved):
LGTM. PR description looks stale, do update before you merge.

PR #108, opened 2026-06-29
fix: 32-bit Windows crash, block-cache UAF, and 32-bit AES-NI

mmokhi:
Three independent fixes.

**safe_shared_from_this (greywolf's crash).** It caught `bad_weak_ptr` from `shared_from_this()` on a connection being torn down. On 32-bit mingw the SjLj unwind for that throw landed with a zeroed stack pointer and crashed at `safe_shared_from_this+0xde`. Now uses the noexcept `weak_from_this().lock()`: same empty-on-expiry result, no throw. Reproduced and confirmed fixed on a runner; 64-bit and Linux were never affected.

**block cache UAF.** `get_cna_v*_data` read `m_block_cache` with no lock while `build_block_cache` could reallocate it, freeing the buffer mid-read (Linux mining crash). `m_block_cache_lock` is now a `shared_mutex`: shared lock for readers, unique lock for the growth.

**32-bit AES-NI.** The hardware-AES path was 64-bit only, so x32 builds ran software AES even on AES-NI CPUs ("Hardware AES not detected"). Enabled it for i686. The startup self-test compares HW vs SW hashes and refuses to boot on a mismatch, so it can't mishash.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.h
+++ b/src/blockchain_db/lmdb/db_lmdb.h
@@ -35,6 +35,7 @@
 #include "cryptonote_basic/blobdatatype.h" // for type blobdata
 #include "ringct/rctTypes.h"
 #include <boost/thread/tss.hpp>
+#include <boost/thread/shared_mutex.hpp>
 
 #include <lmdb.h>
 
@@ -519,7 +520,12 @@ private:
 
   std::vector<block_cache_data> m_block_cache;
   std::atomic<uint64_t> m_block_cache_height;
-  mutable epee::critical_section m_block_cache_lock;
+  // shared_mutex: get_cna_v*_data readers take a shared lock for their hot
+  // lookups; build_block_cache takes a unique lock for the reserve/push_back
+  // that can reallocate the vector. A plain mutex was effectively absent on
+  // the read side, so a reallocation from one thread freed the buffer another
+  // thread was mid-read - a use-after-free crash.
+  mutable boost::shared_mutex m_block_cache_lock;
 
 
 #if defined(__arm__)
--- a/src/blockchain_db/lmdb/db_lmdb.h
+++ b/src/blockchain_db/lmdb/db_lmdb.h
@@ -525,6 +525,10 @@ private:
   // that can reallocate the vector. A plain mutex was effectively absent on
   // the read side, so a reallocation from one thread freed the buffer another
   // thread was mid-read - a use-after-free crash.
+  // open/close and the reset path also mutate the vector without this lock:
+  // they run single-threaded at init/shutdown, before any reader exists or
+  // after they are gone. set_expected_min_height can run while readers are
+  // live, so it does take the lock.
   mutable boost::shared_mutex m_block_cache_lock;
 
 
--- a/contrib/epee/include/net/abstract_tcp_server2.inl
+++ b/contrib/epee/include/net/abstract_tcp_server2.inl
@@ -129,15 +129,12 @@ PRAGMA_WARNING_DISABLE_VS(4355)
   template<class t_protocol_handler>
   boost::shared_ptr<connection<t_protocol_handler> > connection<t_protocol_handler>::safe_shared_from_this()
   {
-    try
-    {
-      return connection<t_protocol_handler>::shared_from_this();
-    }
-    catch (const boost::bad_weak_ptr&)
-    {
-      // It happens when the connection is being deleted
-      return boost::shared_ptr<connection<t_protocol_handler> >();
-    }
+    // Returns empty while the connection is being deleted, same as catching
+    // bad_weak_ptr from shared_from_this() but without the throw. On 32-bit mingw
+    // the throwing path crashed: the SjLj exception unwind landed in the catch
+    // with a corrupted (zero) stack pointer. weak_from_this().lock() is noexcept
+    // and uses the same atomic acquire, so every "if(!self)" caller still works.
+    return this->weak_from_this().lock();
   }
   //---------------------------------------------------------------------------------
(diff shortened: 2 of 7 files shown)


old-greywolf:
on a system with the 32-bit Windows 10 Pro 21H2 the daemon was launched using the nerva-windows-x32 artifacts above, and the first display output was:  Hardware AES (AES-NI / ARMv8 Crypto Extensions) was not detected. Falling back to the software AES path; hashing will be substantially slower. 
then the daemon continued and displayed the nerva ASCII art and the version v0.2.3.0: Legacy Reborn, and continued.  issuing a status to the screen resulted in the normal report with the version "software version 0.2.3.0-a9692685e".

mmokhi:
found it. the crash was the p2p teardown throwing bad_weak_ptr through 32-bit mingw's exception unwinder, which trashes the stack pointer. that's your safe_shared_from_this+0xde APPCRASH exactly, 32-bit windows only (64-bit and linux are fine).

this PR makes that path non-throwing. repro'd it on a runner, patched survives where the old build dies in ~26s. also fixed the "hardware AES not detected" line while i was in there.

feel free to test and validate

old-greywolf:
retested daemon with updated artifacts binary software version 0.2.3.0-292b59380, and the AES-NI warning is not printed to the terminal.  however, there are messages about SSL handshake failure:
INFO	SSL handshake failed, connection dropped: wrong version number (SSL routines)
2026-06-30 22:14:52.290
ERROR	SSL handshake failed on an autodetect connection, reconnecting without SSL

mmokhi:
> retested daemon with updated artifacts binary software version 0.2.3.0-292b59380, and the AES-NI warning is not printed to the terminal. however, there are messages about SSL handshake failure: INFO SSL handshake failed, connection dropped: wrong version number (SSL routines) 2026-06-30 22:14:52.290 ERROR SSL handshake failed on an autodetect connection, reconnecting without SSL

Wonderful! :tada: 

Yes it's all good! 
Those SSL lines are the analytics ping, it posts to a plain http:// endpoint, so the client's SSL autodetect probe fails with "wrong version number" and falls back to plaintext. Launching with `--no-analytics` makes it go away :)
I also pushed a "fix" here so the client skips the SSL probe on http:// urls so that silences it as well...

The ultimate fix for this is to point the analytics.nerva.one TXT record at https://api.nerva.one/ IMHO but not very much in the scope of the PR

src/blockchain_db/lmdb/db_lmdb.h
```diff
@@ -519,7 +520,12 @@ class BlockchainLMDB : public BlockchainDB
 
   std::vector<block_cache_data> m_block_cache;
   std::atomic<uint64_t> m_block_cache_height;
-  mutable epee::critical_section m_block_cache_lock;
+  // shared_mutex: get_cna_v*_data readers take a shared lock for their hot
+  // lookups; build_block_cache takes a unique lock for the reserve/push_back
+  // that can reallocate the vector. A plain mutex was effectively absent on
+  // the read side, so a reallocation from one thread freed the buffer another
+  // thread was mid-read - a use-after-free crash.
+  mutable boost::shared_mutex m_block_cache_lock;
```
sn1f3rt, line 532: nit: `set_expected_min_height()` (`reserve`), `close()`, and the reset path (`clear()`) also mutate `m_block_cache` without taking this lock. They only run at init/shutdown so there is no live race today, but a short invariant comment (or taking the unique lock there too) would keep a future caller from reintroducing the use-after-free.

sn1f3rt (approved):
Approving. Verified the three fixes hold up:

- `safe_shared_from_this`: `weak_from_this().lock()` is a valid noexcept replacement for the throwing `shared_from_this()`/`catch(bad_weak_ptr)` pair (the connection derives from `boost::enable_shared_from_this`, which provides `weak_from_this()`), same empty-on-expiry result.
- block-cache UAF: the shared/unique split makes the v2/v4/v5 readers race-free. `build_block_cache` runs and releases its unique lock before the shared read, and the vector only ever grows, so the indices read under the shared lock stay valid.
- 32-bit AES-NI: `-maes` is applied to 32-bit x86 (the `CMakeLists.txt` gate is not `BUILD_64`-conditional), so `slow-hash-hw.c` compiles, and the startup self-test guards against a mismatch.

One follow-up for the merge-forward into `hard-fork-13`: `get_cna_v6_data` (the post-HF13 mining path, which does not exist on master) reads `m_block_cache` in a loop after `build_block_cache(height)` with no lock, so the exact use-after-free this PR fixes stays open there. It needs the same `boost::shared_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);` line right after its `build_block_cache` call. Not blocking this PR since v6 is not on master.

src/blockchain_db/lmdb/db_lmdb.h
mmokhi: good point. went with the split: set_expected_min_height can run while readers are live so it now takes the unique lock; open/close/reset stay lockless since they're single-threaded init/shutdown, with a comment at the lock saying so.

sn1f3rt (approved):
Re-approving on eddae053. The inline nit is addressed: `set_expected_min_height` now takes the unique lock before the `reserve` that can reallocate, and the header documents why open/close/reset stay lock-free (single-threaded init/shutdown) while this path does not. Correct and no deadlock (called from `Blockchain::init` holding no other lock). The three core fixes are unchanged from the prior approval.

Reminder for later, not blocking here: when this merges forward into `hard-fork-13`, `get_cna_v6_data` still needs the same `boost::shared_lock` after its `build_block_cache` call - v6 is not on master, so it can only be fixed on that branch.

PR #110, opened 2026-07-02
fix(hf13): try harder for huge pages and stop degrading silently

mmokhi:
Follow-up to the hashrate findings on discord. allocate_hugepage() fell back to plain malloc silently when huge pages were unavailable, so v13 mining just looked slow until the next reboot with nothing in the log explaining why.

What it does now:
- Linux: hugetlb, then a THP hint on regular pages (usually still 2 MB pages on default distro settings), then malloc
- Windows: large-page requests rounded up to GetLargePageMinimum() (the 1 MB pad and the salt could never qualify before), plus one retry after trimming our working set when the first attempt fails while SeLockMemory is held
- FreeBSD: superpage-aligned mapping, prefaulted so the kernel can promote it
- macOS/other BSDs: unchanged, just labeled honestly

New one-shot `nervad --setup-large-pages` for Windows: grants "Lock pages in memory" to your user through the LSA policy (works on Home which has no gpedit), run once as administrator and relogon. Explicit on purpose, the daemon never touches system policy on its own. Linux and FreeBSD need no setup.

Miner threads log which page tier their scratchpad landed on, and there is one visible warning per mining session (user log category, shows at every log level) when it ends up on plain pages, with instructions on how to fix it.

No consensus impact: page backing changes speed, not hash values. The PoW itself is untouched.

Follow-up to the affinity findings: new opt-in `--mining-affinity` pins one mining thread per physical core when mining starts (SMT siblings skipped), and keeps the whole set inside a single L3 group when the thread count fits, so the other chiplet stays with your apps. Per-thread, no privileges, dies with the process, off by default. Unreadable topology means it just does not pin.

### DIFF
--- a/src/cryptonote_basic/miner.cpp
+++ b/src/cryptonote_basic/miner.cpp
@@ -639,8 +639,8 @@ namespace cryptonote
       CRITICAL_REGION_END();
       if (tier < good_tier && template_version >= 13 && !m_slow_pages_warned.exchange(true))
         MGUSER_YELLOW("Mining is running on normal memory pages, hashrate will be lower. "
-            "Huge/large pages were not available (Windows: enable \"Lock pages in memory\" for your user; "
-            "Linux: set vm.nr_hugepages or leave transparent hugepages enabled). "
+            "Windows: run 'nervad --setup-large-pages' once as administrator, then log out and back in. "
+            "Linux: set vm.nr_hugepages or leave transparent hugepages enabled. "
             "Freeing memory or rebooting, then restarting mining, retries the allocation.");
     }
     block b;
--- a/src/cryptonote_basic/miner.cpp
+++ b/src/cryptonote_basic/miner.cpp
@@ -114,6 +114,7 @@ namespace cryptonote
     m_pbc(pbc),
     m_height(0),
     m_threads_active(0),
+    m_slow_pages_warned(false),
     m_pausers_count(0),
     m_threads_total(0),
     m_donate_percent(MINING_DEFAULT_DONATION_LEVEL),
@@ -480,6 +481,8 @@ namespace cryptonote
 
     boost::interprocess::ipcdetail::atomic_write32(&m_stop, 0);
     boost::interprocess::ipcdetail::atomic_write32(&m_thread_index, 0);
+    // page situation can change between sessions, warn again if still bad
+    m_slow_pages_warned = false;
     set_is_background_mining_enabled(do_background);
     set_ignore_battery(ignore_battery);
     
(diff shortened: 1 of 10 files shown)


ngeojiajun (approved):

src/cryptonote_basic/miner.cpp
```diff
@@ -29,6 +29,13 @@
 //
 // Parts of this file are originally copyright (c) 2012-2013 The Cryptonote developers
 
+#ifdef _WIN32
+// processor-group topology types need Win7 headers; the two functions are
+// resolved at runtime below so older Windows still runs, it just does not pin
+#undef _WIN32_WINNT
+#define _WIN32_WINNT 0x0601
+#endif
```
sn1f3rt: Force-bumping `_WIN32_WINNT` in this TU is a latent ODR/ABI hazard on the MinGW build.

The MinGW build sets `_WIN32_WINNT=0x0501` globally (`CMakeLists.txt:796`). The bump to `0x0601` is genuinely needed for the processor-group topology structs, but it means `miner.cpp` compiles boost.thread and the out-of-line members of class `miner` at a different Windows API level than every other TU that includes `miner.h` -- same class, two API levels, on the primary Windows release path. It also raises only `_WIN32_WINNT`, not the `WINVER`/`NTDDI_VERSION` trio, which is inconsistent.

Safer approaches, in order of preference:
1. Locally declare the handful of structs/typedefs the enumeration needs (what XMRig does) and drop the global bump entirely, or
2. Move `enumerate_cores`/`pin_this_thread` into their own small `.cpp` so the boost-heavy TU keeps the project-wide API level.

If you keep the bump, at least raise `WINVER`/`_WIN32_WINNT`/`NTDDI_VERSION` together and guard it: `#if !defined(_WIN32_WINNT) || _WIN32_WINNT < 0x0601`.

src/cryptonote_basic/miner.cpp
```diff
@@ -480,6 +648,19 @@ namespace cryptonote
 
     boost::interprocess::ipcdetail::atomic_write32(&m_stop, 0);
     boost::interprocess::ipcdetail::atomic_write32(&m_thread_index, 0);
+    // page situation can change between sessions, warn again if still bad
+    m_slow_pages_warned = false;
+    m_affinity_plan.clear();
+    if (m_mining_affinity)
+      m_affinity_plan = plan_mining_affinity(m_threads_total);
```
sn1f3rt: `--mining-affinity` only pins one thread when the thread count is autodetected -- which is the default when `--mining-threads` isn't set.

The plan is built here from `m_threads_total`, which is `1` at this point under autodetect (`start()` sets it to `1` at lines 459/462). Autodetect then ramps up more workers over the next few seconds, but `m_affinity_plan` is never rebuilt, so every worker with `th_local_index >= 1` fails the `th_local_index < m_affinity_plan.size()` guard in `worker_thread()` and is left unpinned. A user who enables the flag without an explicit thread count gets `Pinning 1 mining thread(s)` and almost none of the intended single-L3 locality -- which is the whole point of the feature.

Options: rebuild the plan once autodetect settles on `m_threads_total` (around line 320), or document that affinity requires an explicit `--mining-threads`.

src/cryptonote_basic/miner.cpp
mmokhi: good catch, went with your option 1: the bump is gone, the file now carries local replicas of the four topology structs (fixed windows ABI) and keeps resolving the two functions from kernel32 at runtime. so the TU compiles at the project-wide 0x0501 like everything else and old windows still runs, it just does not pin.

src/cryptonote_basic/miner.cpp
mmokhi: right, the plan was built once at start() and autodetect outgrew it. plan building moved to a helper that also runs on every autodetect respawn, so each ramp step and the final count get pinned correctly. thanks, this one was a real hole.

PR #115, opened 2026-07-07
fix: anchor quicksync to checkpoints and harden the loader

### LINKED ISSUE #114: security: quicksync trust gaps - unanchored PoW-skip, parser OOM, and mismatch fall-through
## Summary

The quicksync trust system (`src/checkpoints/quicksync.{cpp,h}`, consumed in `src/cryptonote_core/blockchain.cpp`) skips proof-of-work for blocks whose hash is listed in a downloaded `quicksync.raw` file. The file is trusted with no cryptographic anchor and no bounds, which leads to three defects found during a full-tree security review.

## C1 (Critical) - PoW-skip is not anchored to the checkpoints and is unbounded above the top checkpoint

`quicksync::load()` accepts the file gated only by a 4-byte magic - no signature, no hash, and no cross-check against the compiled-in checkpoints. During block verification any block whose id matches the file's hash for its height has PoW skipped, and that skip is:

- not bounded to the highest hardcoded checkpoint (currently height 4,290,000), so entries above it are trusted with no checkpoint anchor and no PoW, and
- not gated by `--fast-block-sync`, so even a node run with `--fast-block-sync 0` still gets PoW skipped for quicksync-matched blocks.

Impact: a poisoned `quicksync.raw` (hostile mirror / MITM) whose entries cover heights above the top checkpoint, combined with matching fabricated blocks from a hostile/eclipsing peer, lets a syncing node accept a forged chain extension with no valid PoW. Below the top checkpoint the exposure is bounded by the sparse checkpoints plus `prev_id` chaining; the dangerous window is strictly above the highest checkpoint.

## H1 (High) - integer underflow and unchecked reads in the parser -> OOM on load

`quicksync::load()` reads `m_min`/`m_max` without checking the reads, computes `count = m_max - m_min` in 32-bit (underflows to ~4.29e9 when `m_min > m_max`), and never checks per-hash reads or bounds the count against the actual file size. A malformed/hostile file makes the daemon allocate billions of map entries (>100 GB) and OOM at startup, or spin on a truncated file.

## M2 (Medium) - a quicksync hash mismatch is not distinguished from "absent"

`quicksync::check_block` returns `false` both when a height is absent and when the stored hash mismatches, and the caller does not distinguish them. A block covered by quicksync but with the wrong hash therefore falls through to the assume-valid PoW-skip below `ASSUME_VALID_HEIGHT` and is committed to the main chain.

## Affected files

- `src/checkpoints/quicksync.cpp`, `src/checkpoints/quicksync.h`
- `src/cryptonote_core/blockchain.cpp` (block verification path in `handle_block_to_main_chain`, and `set_quicksync`)

## Scope

C1/H1/M2 share the same root cause (the file is trusted with no anchor and no bounds) and should be addressed together. The related `assume_valid` fast-sync behavior below the top checkpoint is out of scope here - it remains checkpoint-anchored, so a fake chain dead-ends at the next checkpoint. Other review findings are tracked separately.

sn1f3rt:
Closes #114.

Anchors the quicksync trust system to the compiled-in checkpoints and hardens the loader. Three defects, one root cause: the downloaded `quicksync.raw` was trusted with no cryptographic anchor and no bounds.

## Changes

**`quicksync::load()` hardening (H1)**
- Check every header/hash read succeeds; reject `min > max`.
- Compute the entry count in 64-bit and bound it by the actual file size before allocating, so a corrupt/hostile header can't drive an unbounded allocation (previously `count = m_max - m_min` underflowed to ~4.29e9 and OOM'd).
- Store 64-bit heights (was a `uint32_t` counter that could wrap).

**Checkpoint anchoring (C1)**
- New `quicksync::check_against_checkpoints()` compares every quicksync hash at a hardcoded-checkpoint height and rejects the whole file on any conflict.
- `Blockchain::set_quicksync()` is now a validating setter that drops a conflicting file instead of trusting it, and records the checkpoint height it validated against (`m_quicksync_max_height`). This runs during `handle_command_line`, before DNS/JSON checkpoints load, so it captures the hardcoded checkpoint max.

**Consume path (C1 + M2)** (`handle_block_to_main_chain`)
- Quicksync may only skip PoW when `--fast-block-sync` is on and `height <= m_quicksync_max_height` (the checkpoint height it was anchored against). Above that the file is unanchored, so PoW is always required. Because the ceiling is captured before DNS/JSON checkpoints load, runtime checkpoints can't stretch the trusted range past what was validated. Also fixes quicksync ignoring `--fast-block-sync 0`.
- A quicksync entry that is present but whose hash mismatches now falls back to full PoW (with a warning) rather than being trusted or hard-failing: a real block still verifies, a forged one still fails PoW, and a wrong/corrupt file only loses the skip-speedup instead of stalling the node.

Net effect: quicksync becomes a checkpoint-anchored PoW-skip within the validated range, and defers to PoW everywhere else.

## Testing

- `make obj_checkpoints` and `make obj_cryptonote_core` build clean; all static libraries link. (The only unresolved symbol in a full `nervad` link is the pre-existing `_IOPSGetTimeRemainingEstimate` from `miner.cpp` on native macOS, unrelated to this change.)
- Unit-tested `quicksync::load()`, `check_against_checkpoints()`, and `check_block()` via a standalone harness linked against `libcheckpoints.a` (17/17): parser rejects `min > max` / truncated / oversized-count / short-body files with no OOM, the checkpoint cross-check accepts matching and rejects conflicting files, and `check_block` distinguishes absent / matching / mismatching. (See PR comment for the full case list.)
- Reviewed against the export format (`quicksync_file.cpp` writes `[start, stop]` inclusive): the file-size bound has one hash of slack versus `count = m_max - m_min`, so a valid file is never rejected.
- Only one caller of the changed `check_block` signature exists (the consume path). testnet/stagenet have no compiled-in checkpoints, so `m_quicksync_max_height` is 0 and quicksync is inert there (safe).

## Scope

C1/H1/M2 only. The `assume_valid` fast-sync behavior below the HF13 height is unchanged and remains checkpoint-anchored. Other review findings are tracked separately.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -3570,27 +3570,26 @@ leave:
   crypto::hash proof_of_work;
   memset(proof_of_work.data, 0xff, sizeof(proof_of_work.data));
 
-  // Both skips below are bounded to the range the hardcoded checkpoints pin, and
-  // gated by --fast-block-sync; FAKECHAIN inherits the mainnet config but never
+  // Gated by --fast-block-sync; FAKECHAIN inherits the mainnet config but never
   // reaches that height.
   const uint64_t assume_valid_height = cryptonote::get_config(m_nettype).ASSUME_VALID_HEIGHT;
   const bool assume_valid = m_fast_sync && m_nettype != FAKECHAIN && assume_valid_height != 0 && blockchain_height < assume_valid_height;
 
-  // Above the highest checkpoint the file is unanchored, so PoW is always required.
-  const bool quicksync_active = m_fast_sync && m_nettype != FAKECHAIN && blockchain_height <= m_checkpoints.get_max_height();
+  // Quicksync is trusted only up to the checkpoint height it was anchored against
+  // (m_quicksync_max_height, captured at set_quicksync before any DNS/JSON checkpoints);
+  // above that it is unanchored, so PoW is always required.
+  const bool quicksync_active = m_fast_sync && m_nettype != FAKECHAIN && blockchain_height <= m_quicksync_max_height;
   bool quicksync_has_block = false;
   const bool quicksync_match = m_quicksync.check_block(blockchain_height, id, quicksync_has_block);
-
-  // A present-but-mismatching entry must fail hard, not fall through to the assume-valid skip.
-  if (quicksync_active && quicksync_has_block && !quicksync_match)
-  {
-    MERROR_VER("Block with id: " << id << " at height " << blockchain_height << " does not match the quick sync hash");
-    bvc.m_verifivation_failed = true;
-    goto leave;
-  }
-
   const bool quicksync_verified = quicksync_active && quicksync_match;
-  if (!quicksync_verified && !assume_valid)
+
+  // A present-but-mismatching entry could be a forged block or a wrong/corrupt file;
+  // let PoW decide rather than trusting the file or stalling the node on a bad file.
+  const bool quicksync_conflict = quicksync_active && quicksync_has_block && !quicksync_match;
+  if (quicksync_conflict)
+    MWARNING("Block with id: " << id << " at height " << blockchain_height << " does not match the quick sync hash; verifying PoW");
+
+  if ((!quicksync_verified && !assume_valid) || quicksync_conflict)
   {
     get_block_longhash(m_hash_context, this, bl, proof_of_work, blockchain_height);
 
@@ -4116,7 +4115,11 @@ void Blockchain::set_quicksync(quicksync&& qs)
     MERROR("Quick sync data conflicts with hardcoded checkpoints; ignoring quick sync file");
     return;
   }
-  m_quicksync = qs;
+  // Pin the trusted range to the checkpoints just validated against. This runs
+  // before DNS/JSON checkpoints load, so it is the hardcoded checkpoint max and
+  // can't be stretched at runtime.
+  m_quicksync_max_height = m_checkpoints.get_max_height();
+  m_quicksync = std::move(qs);
 }
 //------------------------------------------------------------------
 bool Blockchain::cleanup_handle_incoming_blocks(bool force_sync)
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -3570,22 +3570,26 @@ leave:
   crypto::hash proof_of_work;
   memset(proof_of_work.data, 0xff, sizeof(proof_of_work.data));
 
-  // An earlier revision skipped PoW for the entire checkpoint zone unconditionally
-  // (an unanchored, dangerous skip) and that was removed. The assume-valid skip
-  // below is the deliberate, bounded replacement: PoW is dropped only for blocks
-  // beneath a hardcoded checkpoint that already pins them by hash, and only during
-  // flag-gated fast sync.
-  // FIXME: height parameter is not used...should it be used or should it not
-  // be a parameter?
-  // validate proof_of_work versus difficulty target
-
-  // Skip PoW for deep historical blocks below ASSUME_VALID_HEIGHT during sync
-  // (anchored by a checkpoint). Gated by --fast-block-sync; FAKECHAIN excluded
-  // as it inherits the mainnet config but never reaches that height.
+  // Both skips below are bounded to the range the hardcoded checkpoints pin, and
+  // gated by --fast-block-sync; FAKECHAIN inherits the mainnet config but never
+  // reaches that height.
   const uint64_t assume_valid_height = cryptonote::get_config(m_nettype).ASSUME_VALID_HEIGHT;
   const bool assume_valid = m_fast_sync && m_nettype != FAKECHAIN && assume_valid_height != 0 && blockchain_height < assume_valid_height;
 
-  const bool quicksync_verified = m_quicksync.check_block(blockchain_height, id);
+  // Above the highest checkpoint the file is unanchored, so PoW is always required.
+  const bool quicksync_active = m_fast_sync && m_nettype != FAKECHAIN && blockchain_height <= m_checkpoints.get_max_height();
+  bool quicksync_has_block = false;
+  const bool quicksync_match = m_quicksync.check_block(blockchain_height, id, quicksync_has_block);
+
+  // A present-but-mismatching entry must fail hard, not fall through to the assume-valid skip.
+  if (quicksync_active && quicksync_has_block && !quicksync_match)
+  {
+    MERROR_VER("Block with id: " << id << " at height " << blockchain_height << " does not match the quick sync hash");
+    bvc.m_verifivation_failed = true;
+    goto leave;
+  }
+
+  const bool quicksync_verified = quicksync_active && quicksync_match;
   if (!quicksync_verified && !assume_valid)
   {
     get_block_longhash(m_hash_context, this, bl, proof_of_work, blockchain_height);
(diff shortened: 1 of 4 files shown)


sn1f3rt:
Unit-tested the changed quicksync functions (`load()`, `check_against_checkpoints()`, `check_block()`) via a standalone harness linked against the built `libcheckpoints.a`, driving crafted files. 17/17 passed.

**H1 - `load()` parser**
- valid file loads
- `min > max` rejected
- truncated header rejected
- truncated body rejected
- huge count + tiny file rejected (no OOM; `min=0, max=0xFFFFFFFF` returns immediately)
- wrong magic rejected
- empty file rejected
- `min == max` loads as empty set
- exact-size loads
- one-byte-short rejected

**C1 - `check_against_checkpoints()`**
- hash matching a checkpoint -> accepted
- hash conflicting with a checkpoint -> rejected
- checkpoint outside the file's range -> accepted (no-op)
- empty checkpoints -> accepted

**M2 - `check_block()` tri-state**
- present + matching -> `true`, `is_present=true`
- present + mismatch -> `false`, `is_present=true`
- absent -> `false`, `is_present=false`

Scope: this covers the functions in `libcheckpoints.a`. The consume-path wiring in `blockchain.cpp` (`height <= get_max_height()` bound, `--fast-block-sync` gate, mismatch hard-fail) is covered by compile + inspection, not runtime.

old-greywolf:
using nerva-macos-armv8, full sync from 0 was 2 hours 55 minutes.  deleted DB and then used quicksync from 0 and time was 41 minutes.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3570,22 +3570,26 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
...
   const bool assume_valid = m_fast_sync && m_nettype != FAKECHAIN && assume_valid_height != 0 && blockchain_height < assume_valid_height;
 
-  const bool quicksync_verified = m_quicksync.check_block(blockchain_height, id);
+  // Above the highest checkpoint the file is unanchored, so PoW is always required.
+  const bool quicksync_active = m_fast_sync && m_nettype != FAKECHAIN && blockchain_height <= m_checkpoints.get_max_height();
+  bool quicksync_has_block = false;
+  const bool quicksync_match = m_quicksync.check_block(blockchain_height, id, quicksync_has_block);
+
+  // A present-but-mismatching entry must fail hard, not fall through to the assume-valid skip.
+  if (quicksync_active && quicksync_has_block && !quicksync_match)
```
mmokhi: To me it seems like quicksync is an operator-supplied file, not compiled in like the hardcoded checkpoints, so a present-but-mismatching entry might have two causes we can't really tell apart: a forged block (its real hash is the quicksync one), or a corrupt/tampered quicksync file at a non-checkpoint height (the block is real, the file is wrong). Hard-failing seems right for the first, but for the second I think it would reject the real block from every peer and stall sync until the operator deletes the file.

Since we can't tell them apart, would it be safer to let PoW adjudicate? i.e. on mismatch, force full PoW rather than failing the block (and not let the assume-valid skip fire either). A real block would still pass and be accepted, a forged one would still fail PoW and be rejected as today, and a corrupt file would just cost the skip-speedup instead of halting the node. Something like:

```cpp
const bool quicksync_conflict = quicksync_active && quicksync_has_block && !quicksync_match;
const bool quicksync_verified = quicksync_active && quicksync_match;
if ((!quicksync_verified && !assume_valid) || quicksync_conflict)
{
    // existing get_block_longhash + PoW check
}
```

with a warning logged on conflict so a bad file is still loud. Might be missing something though.

src/cryptonote_core/blockchain.cpp
```diff
@@ -3570,22 +3570,26 @@ bool Blockchain::handle_block_to_main_chain(const block& bl, const crypto::hash&
...
-  // as it inherits the mainnet config but never reaches that height.
+  // Both skips below are bounded to the range the hardcoded checkpoints pin, and
+  // gated by --fast-block-sync; FAKECHAIN inherits the mainnet config but never
+  // reaches that height.
   const uint64_t assume_valid_height = cryptonote::get_config(m_nettype).ASSUME_VALID_HEIGHT;
   const bool assume_valid = m_fast_sync && m_nettype != FAKECHAIN && assume_valid_height != 0 && blockchain_height < assume_valid_height;
 
-  const bool quicksync_verified = m_quicksync.check_block(blockchain_height, id);
+  // Above the highest checkpoint the file is unanchored, so PoW is always required.
+  const bool quicksync_active = m_fast_sync && m_nettype != FAKECHAIN && blockchain_height <= m_checkpoints.get_max_height();
```
mmokhi: `get_max_height()` is the top of all of `m_points`, which includes JSON/DNS checkpoints added after init, not only the hardcoded ones. `check_against_checkpoints()` ran at `set_quicksync` time against the hardcoded set only (DNS/JSON not loaded yet), so this PoW-skip zone can extend past the range the file was actually anchored against. Still safe via the checkpoint check at ~L3607 plus block linkage, but is that intended, or should the zone be bounded to the hardcoded max?

src/cryptonote_core/blockchain.cpp
```diff
@@ -4103,7 +4107,17 @@ void Blockchain::set_enforce_dns_checkpoints(bool enforce_checkpoints)
...
+//------------------------------------------------------------------
+void Blockchain::set_quicksync(quicksync&& qs)
+{
+  // Reject an unanchored/forged file rather than trusting it to skip PoW.
+  if (!qs.check_against_checkpoints(m_checkpoints))
+  {
+    MERROR("Quick sync data conflicts with hardcoded checkpoints; ignoring quick sync file");
+    return;
+  }
+  m_quicksync = qs;
```
mmokhi:
```suggestion
  m_quicksync = std::move(qs);
```
`qs` is an rvalue-ref param holding the full hash map (potentially millions of entries), isn't it better to move it than copy?

src/cryptonote_core/blockchain.cpp
sn1f3rt: Good call, agreed. Changed in 35128c9: a present-but-mismatching entry now falls back to full PoW instead of hard-failing, with a warning logged. A real block still verifies, a forged one still fails PoW, and a corrupt file just loses the skip-speedup instead of stalling the node. Used your exact condition: `(!quicksync_verified && !assume_valid) || quicksync_conflict`.

src/cryptonote_core/blockchain.cpp
sn1f3rt: Right. Fixed in 35128c9 by capturing the checkpoint max in set_quicksync (m_quicksync_max_height) and bounding the skip to that, instead of get_max_height() at consume time. set_quicksync runs in handle_command_line (core::init:474), before DNS/JSON checkpoints load via update_checkpoints (core::init:697), so the captured value is the hardcoded checkpoint max and runtime checkpoints cannot stretch the trusted range.

src/cryptonote_core/blockchain.cpp
sn1f3rt: Done in 35128c9.

src/cryptonote_core/blockchain.cpp
mmokhi: Nice, that reads much safer to me. Thanks for taking it.

src/cryptonote_core/blockchain.cpp
mmokhi: Great! thanks

src/cryptonote_core/blockchain.cpp
mmokhi: Perfect, thanks.

mmokhi (approved):
LGTM,
Great work! thanks

ngeojiajun (approved):

Issue #116, opened 2026-07-09
HF14: CLSAG + Bulletproofs+, and CryptoNight-Adaptive v7


mmokhi:
Tracking issue for hard fork 14.

Per the roadmap (https://nerva.one/potential-future-nerva-hard-forks/), HF14
upgrades the transaction format: CLSAG ring signatures replace MLSAG (about
25% smaller signatures, faster verification) and Bulletproofs+ replace
Bulletproofs (about 6% smaller, a few percent faster to verify). The same
fork carries the PoW change, CryptoNight-Adaptive v7: the memory-hard work
becomes a strictly serial chase over a 24 MB per-nonce buffer that the walk
fills from the seed and writes back to as it goes. Each hop takes its address
operands from the per-nonce program and steps through them by a stride drawn
from the value just loaded, and the chase segment lengths come from the seed
too, so the memory pattern is program-driven the way v6's was while the access
count per pass stays fixed and no seed yields a cheaper hash.

Measured, on real hardware where possible:
- per-core gap 1.37x: 570 ms per hash on an aarch64 Cortex-A57 core at 1.43 GHz
  against 416 ms on a 4.8 GHz x86-64 core, medians of 7, both idle. v6 measured
  the same way on the same two machines is 4.86x (93.3 ms against 19.2 ms), and
  the clock ratio is 3.36x: v7 lands well under it, v6 above it.
- GPU 1.66x a 14-thread x86-64 box on a 16 GB card, against 3.07x for a chase
  whose addresses came from a fixed formula, and 4.66x for v6. All measured on
  the same card in one run.
- ASIC: cannot be measured, no silicon. Structurally the barrier is the
  per-nonce program governing the addresses plus buffer residency plus DRAM
  latency. v6 remains the reference point here rather than something v7 beats.
- verify cost is the axis where v7 loses to v6: about 410 ms per hash on a
  4.8 GHz core and 572 ms on the Cortex-A57, against roughly 14 ms for v6,
  and 25 MB per thread against 9.25 MB. That is the price of binding to DRAM
  latency and it is deliberate.

Implementation is complete (PR #117). Placeholder heights are in: mainnet 14 at
4400000, testnet 12/13/14 at 700/800/1000, stagenet 14 at 1000. The testnet
heights need a testnet started fresh from genesis, since 12 and 13 move down
from 2000/2100: wipe the testnet data directory before the run.

Remaining before the mainnet height is set:
[...]

PR #117, opened 2026-07-09
Hard fork 14: CLSAG + Bulletproofs+, CryptoNight-Adaptive v7

mmokhi:
Implements HF14: the privacy upgrade from the roadmap (CLSAG +
Bulletproofs+) and the CryptoNight-Adaptive v7 PoW change, activating
together at fork 14. Tracking issue #116 - its validation gates remain
open past this merge, so it should stay open.

For users: transactions get about 25% smaller and verify faster, same
privacy. For miners: per-core hashrate evens out across consumer
hardware, small boards stay first-class. Nothing changes on any network until fork heights are set, and none is
yet; heights are chosen network by network (testnet first) per the
validation gates in #116.

Only rct type 7 (CLSAG + Bulletproofs+) activates for new transactions.
Type 6 (CLSAG with the older Bulletproofs) is fully implemented and
verifiable but rejected at every height - it is the intermediate the
port went through, kept so the CLSAG and Bulletproofs+ steps stay
auditable on their own.

## Commits

- `fix(serialization): read_hex string length accessor`
- `fix(ringct): overflow guards and initialized type byte in rctSig`
- `feat(ringct): CLSAG signatures, ported from Monero`
- `feat(ringct): Bulletproofs+ range proofs, ported from Monero`
- `feat(consensus): verify and gate CLSAG + Bulletproofs+ at HF14`
- `refactor(cryptonote_core): adopt tx_verification_utils for input proof verification`
- `feat(wallet): create type-7 (CLSAG + Bulletproofs+) transactions at HF14`
- `feat(hf14): CryptoNight-Adaptive v7, per-nonce mutable-buffer chase`
- `test(hf14): out-of-tree check harness for the HF14 port` (plus two
  `fixup!` commits from the review, kept unsquashed so the fixes stay
  reviewable against their comments)

## Deviations by design

Ported against Monero master 52f02f2. RCTTypeCLSAG = 6 and
RCTTypeBulletproofPlus = 7 (Monero's 5/6 are already taken by our
on-chain history). Deliberately not included: upstream's multisig rework
(legacy multisig is disabled instead, see below), the input verification
cache (the file layout lands without it), view tags and ring size 16
(HF15 per the roadmap). Multisig setup is disabled - the legacy
implementation is unaudited; existing multisig wallets warn and can spend
until the fork. Hardware-device wallets cannot sign HF14 transactions
(pre-CLSAG Ledger protocol).

## Verification

- Every ported function diffs at zero lines against upstream, or with
  each deviation named; the two bulletproofs_plus files are
  byte-identical copies.
- Monero's CLSAG and bulletproofs_plus test batteries ported and
  passing, including tamper and torsion cases. These, plus checks for
  the Nerva-specific glue the upstream diff cannot vouch for (the 6/7
  renumbering through all three encoders, the fee and weight
  arithmetic, CNA v7 against a standalone reference), are now in-tree
  under `contrib/hf14checks` (review ask; details in #116).
- Cross-implementation interop both directions, 24 signatures each way:
  balance and range-proof semantics verify as-is and every CLSAG
  cross-verifies; the serialized base is byte-identical except the
  renumbered type byte, which is part of the signed message by design.
- Serialization roundtrips for all three encoders, including pre-change
  archives; legacy behavior bit-stable (type-5 wire encoding and fee
  estimates unchanged).
- Wallet estimators cover real serialized type-7 signatures; the fee
  clawback equals consensus arithmetic.
- CNA v7's per-nonce chase is validated bit-for-bit against a standalone
  reference, and the checkpoint (time-memory) attack on the buffer is
  implemented and comes out slower than just storing it (0.53-0.63x the
  throughput), so the residency requirement holds. The chase issues exactly
  3,145,728 accesses per hash for every seed, and coverage stays 62.6-63.3%
  across 150 seeds, so no seed is cheaper to hash than another. Measured:
  per-core gap 1.37x (570 ms on an aarch64 Cortex-A57 core at 1.43 GHz against
  416 ms on a 4.8 GHz x86-64 core, medians of 7 on idle machines) against 4.86x
  for v6 measured identically, with a 3.36x clock ratio between them; a 16 GB
  GPU at 1.66x a 14-thread x86-64 box against 4.66x for v6, same card one run.
  The ASIC barrier cannot be measured; structurally it rests on the per-nonce
  program driving the addresses, buffer residency and DRAM latency, and v6 is
  the reference point rather than something this beats. Verify cost is where
  v7 loses to v6: ~410 ms per hash on a 4.8 GHz core, 572 ms on the A57,
  against ~14 ms for v6.
- Full builds of every target. The PoW HW/SW self-test now runs on a
  varied seed and asserts the buffer is populated (review catch: the
  original all-zero seed zeroed the fill and exercised under 1% of the
  buffer), and passes on x86_64 and aarch64, all four AES paths.

### DIFF
--- a/src/crypto/cna-vm.h
+++ b/src/crypto/cna-vm.h
@@ -44,6 +44,32 @@
 #define CN_REG_COUNT       8
 #define CN_VM_ITERATIONS   2048
 
+// CNA v7 (HF14): the memory-hard work is a strictly serial chase over a
+// per-nonce buffer that the walk writes back to as it goes. The buffer is
+// filled cheaply from the per-nonce seed, and the chase covers ~63% of it and
+// mutates every cell it touches, so touched cells diverge from the fill and
+// the buffer cannot be regenerated on the fly: a miner is forced to keep the
+// whole thing resident. That caps how many nonces a GPU/ASIC can hold at once
+// (one buffer each), the way v6's 8 MB pad did, while the chase itself stays
+// DRAM-latency-bound so a small board keeps pace with a desktop: one box ~ one
+// vote. Per pass, CN_V7_SEGMENTS segments of [CN_V7_HOPS / CN_V7_SEGMENTS
+// serial hops, then a program slice], each segment gated on a register the
+// program just mutated so the walk cannot advance without executing the
+// per-nonce random program (the GPU/ASIC barrier, same class as v6). All
+// values consensus-critical.
+#define CN_V7_HOPS           1024
+#define CN_V7_SEGMENTS       8
+// Per-nonce chase buffer: 24 MB. Big enough to spill any commodity L3 (so the
+// chase is DRAM-latency-bound, not cache/clock-bound) and to starve a GPU of
+// parallel nonces (VRAM / 24 MB), small enough to keep block verify ~0.3-0.5 s.
+#define CN_V7_BUFFER         (24 * 1024 * 1024)
+// v14 pass count = buffer qwords / hops-per-pass, so the chase touches ~63% of
+// the buffer. That coverage is what forces full residency: below it a
+// checkpoint attacker could store the buffer sparsely and recompute the rest.
+// Verify lands ~0.3-0.5 s/hash, roughly equal on every box since the chase is
+// DRAM-latency-bound. v13 keeps CN_VM_ITERATIONS.
+#define CN_VM_ITERATIONS_V14 (CN_V7_BUFFER / (8 * CN_V7_HOPS))
+
 // Instruction opcodes — kept small so the opcode byte fits in uint8_t.
 typedef enum {
     CN_OP_IADD_RS   = 0,  // r[dst] += r[src] << (shift & 3)
@@ -80,3 +106,16 @@ void cn_vm_generate_program(cn_vm_program_t *prog, const uint8_t seed[32]);
 // loops possible).  Call CN_VM_ITERATIONS times with the same prog and
 // evolving regs to accumulate scratchpad mutations.
 void cn_vm_execute(cn_vm_program_t *prog, uint8_t *scratchpad, uint64_t regs[CN_REG_COUNT]);
+
+// Execute one HF14 pass: CN_V7_SEGMENTS repetitions of [serial chase over
+// the per-nonce buffer, then a slice of prog].  Each hop reads a buffer cell
+// then writes it back (buffer[idx] ^= chain), so touched cells diverge from
+// the fill and the buffer cannot be regenerated on the fly.  buffer_qwords is
+// its length in 64-bit words (>= 1, any length: addressing maps the chain
+// value through a mul128 high word, no power-of-two requirement).  SP_READ
+// consumes the chased values, SP_WRITE mutates the small pad
+// (CN_SCRATCHPAD_MEMORY_V14 bytes, write-hardness only).  *chain_state
+// carries the walk position across passes so the chase stays sequential for
+// the whole hash and cannot be precomputed.  Call CN_VM_ITERATIONS_V14 times.
+void cn_vm_execute_v7(cn_vm_program_t *prog, uint8_t *buffer, uint64_t buffer_qwords,
+                      uint8_t *pad, uint64_t regs[CN_REG_COUNT], uint64_t *chain_state);
--- a/src/crypto/cna-vm.h
+++ b/src/crypto/cna-vm.h
@@ -57,8 +57,21 @@
 // program just mutated so the walk cannot advance without executing the
 // per-nonce random program (the GPU/ASIC barrier, same class as v6). All
 // values consensus-critical.
+//
+// The hop itself takes its address operands from the program (regs[src] + imm +
+// chain, as v6 did) and steps through the operand source by a stride derived
+// from the value just loaded, so the sequence of operands is per-nonce and
+// value-dependent rather than a fixed formula. Segment lengths are per-nonce
+// too. Measured against a fixed-loop chase on one GPU: 1.76x a CPU box instead
+// of 3.07x.
 #define CN_V7_HOPS           1024
 #define CN_V7_SEGMENTS       8
+// Per-segment hop counts come from the seed and sum to CN_V7_HOPS, so the
+// access count per pass is fixed while the rhythm differs per nonce. The floor
+// keeps the chased prefix ahead of what the following program slices can
+// consume: slice i consumes at most (CN_PROGRAM_SIZE / CN_V7_SEGMENTS) * (i+1)
+// values, and CN_V7_SEG_MIN * (i+1) is larger.
+#define CN_V7_SEG_MIN        80
 // Per-nonce chase buffer: 24 MB. Big enough to spill any commodity L3 (so the
 // chase is DRAM-latency-bound, not cache/clock-bound) and to starve a GPU of
 // parallel nonces (VRAM / 24 MB), small enough to keep block verify ~0.3-0.5 s.
@@ -94,6 +107,10 @@ typedef struct {
 
 typedef struct {
     cn_vm_instruction_t instructions[CN_PROGRAM_SIZE];
+    // Chase segment lengths for the HF14 pass, from the same seed. They sum to
+    // CN_V7_HOPS, so the access count per pass is identical for every nonce and
+    // no seed yields a cheaper hash; only the rhythm differs.
+    uint16_t seg_hops[CN_V7_SEGMENTS];
 } cn_vm_program_t;
 
 // Generate a deterministic random program from a 32-byte seed.
--- a/src/crypto/cna-vm.h
+++ b/src/crypto/cna-vm.h
@@ -61,7 +61,9 @@
 // The hop itself takes its address operands from the program (regs[src] + imm +
 // chain, as v6 did) and steps through the operand source by a stride derived
 // from the value just loaded, so the sequence of operands is per-nonce and
-// value-dependent rather than a fixed formula. Segment lengths are per-nonce
+// value-dependent rather than a fixed formula. The register index also steps
+// with the hop, so the operand sum cannot be reduced to one precomputed table
+// per segment. Segment lengths are per-nonce
 // too. Measured against a fixed-loop chase on one GPU: 1.76x a CPU box instead
 // of 3.07x.
 #define CN_V7_HOPS           1024
--- a/src/crypto/slow-hash-impl.h
+++ b/src/crypto/slow-hash-impl.h
@@ -334,6 +334,141 @@ void cn_slow_hash_v13(cn_hash_context_t *context, const void *data, size_t lengt
     free(text);
 }
 
+// CNA v7 (HF14): same Keccak/AES framing as v13, but the memory-hard middle is
+// a serial chase over a 24 MB per-nonce buffer that the walk writes back to
+// (cn_vm_execute_v7), not v13's VM over the pad. The pad shrinks to
+// CN_SCRATCHPAD_MEMORY_V14 and carries write-hardness only (v13's 8 MB refill
+// was a clock-bound fast-core amplifier; 256 KB fits every cache, so it costs
+// everyone alike). The pad reuses the first 256 KB of cna_scratchpad, whose
+// 8 MB stays allocated for historical v13 validation anyway. chain_state
+// threads through every pass and folds into a register at the end, so the
+// whole walk is load-bearing.
+void cn_slow_hash_v14(cn_hash_context_t *context, const void *data, size_t length, char *hash,
+                      const uint8_t *seed)
+{
+    uint8_t * const hp_state = context->cna_scratchpad;
+    char * const salt = context->salt;
+    const uint8_t init_size_blk = INIT_SIZE_BLK;
+    const uint32_t init_size_byte = (uint32_t)(init_size_blk * AES_BLOCK_SIZE);
+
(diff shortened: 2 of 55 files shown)


ngeojiajun (approved):

src/crypto/slow-hash.c
```diff
@@ -447,6 +465,20 @@ int cn_slow_hash_self_test(void)
         if (memcmp(hw, sw, HASH_SIZE) != 0) ok = 0;
     }
 
+    /* v14: 256 KB pad + a 24 MB per-nonce buffer the hash fills from the seed
+     * and walks (mutating as it goes). Both paths mutate the pad; resetting
+     * salt/random_values before each call keeps the inputs identical. */
+    {
+        static const uint8_t seed[32] = {0};
```
sn1f3rt: **This self-test runs against an all-zero buffer, so it verifies almost nothing.**

The v14 buffer fill is seeded by `memcpy(&fx, seed, sizeof(uint64_t))` and iterates `fx = (fx ^ (fx >> 29)) * 0xbf58476d1ce4e5b9`. Zero is a fixed point of that map, so a zero seed fills all 3,145,728 qwords with zero. The chase then degenerates: with `chain` small, `high64(chain * N)` is 0, and the walk collapses onto the low indices.

Measured against this branch, built and run locally:

```
v14 with ZERO seed (this seed):         0.0240 s/hash
v14 with REAL seed:                     0.1722 s/hash
buffer qwords ever written, zero seed:  24,495 / 3,145,728  (0.78%)
```

This matters because `cn_slow_hash_self_test` is the only guard against an AES-NI versus software-AES divergence on the v14 path, which would be a split between miners on different CPUs. Any bug in `cn_vm_execute_v7`'s memory handling that only shows on a populated buffer passes this test silently.

Any non-zero seed fixes it. Worth also asserting the fill produced a non-zero buffer, so this cannot regress quietly.

src/crypto/slow-hash-impl.h
```diff
@@ -334,6 +334,141 @@ void cn_slow_hash_v13(cn_hash_context_t *context, const void *data, size_t lengt
...
+    // then mutates ~63% of it as it walks, so its cells can't be regenerated
+    // from the fill and the whole buffer has to stay resident. That residency
+    // is the capacity brake: a GPU/ASIC runs only as many nonces as it has
+    // buffers for. Below that coverage the buffer could be stored sparsely and
+    // recomputed, so the pass count is tied to the buffer size.
+    uint8_t * const buffer = context->cna_v7_buffer;
+    const uint64_t buffer_qwords = CN_V7_BUFFER / sizeof(uint64_t);
+    {
+        uint64_t fx;
+        memcpy(&fx, seed, sizeof(uint64_t));
```
sn1f3rt: **The 24 MB buffer derives from only 8 of the 32 seed bytes.** The remaining 24 reach `cn_vm_generate_program` and nothing else.

Two consequences:

1. Buffer collisions between nonces are findable at roughly 2^32 work, so a miner can cache a fill and reuse it across colliding nonces instead of regenerating. The fill is only about 10 ms of a roughly 180 ms hash, so the saving is around 5 percent. Small, but it does undercut the "per-nonce buffer" framing in the commit message.
2. It is what makes the zero case a total collapse rather than a graceful degradation. See the self-test comment on `slow-hash.c`.

Seeding the fill from all 32 bytes fixes both. Worth doing now rather than later: **this changes the hash output**, so it has to land while the code is still inert. No network has an HF14 height yet, so it is free today. Once testnet has a height it costs a testnet re-fork.

The software-AES twin at line 841 needs the identical change, or the two paths diverge and the self-test above is what would have caught it.

src/crypto/slow-hash.c
```diff
@@ -345,6 +353,11 @@ cn_hash_context_t *cn_hash_context_create(void)
         cn_hash_context_free(ctx);
         return NULL;
     }
+    ctx->cna_v7_buffer_is_mapped = allocate_hugepage(CN_V7_BUFFER, (void **)&(ctx->cna_v7_buffer));
```
sn1f3rt: **Consider allocating this lazily, on first v14 use.**

As written, every hash context grows from about 9.3 MB to about 33.3 MB the moment this ships, for every node, months before HF14 activates and regardless of whether the node ever computes a v14 hash.

That also hits contexts created per call for work unrelated to PoW: `crypto/chacha.h:77` and `:88` (`generate_chacha_key`, on every wallet open and save), `cryptonote_format_utils.cpp:1381` and `:1391` (`encrypt_key` / `decrypt_key`), and `wallet2.cpp:1441`. None of those sites null-check `cn_hash_context_create()`, so an allocation failure is a null deref. That pattern predates this PR, but a 3.6x jump in per-context footprint raises the odds of hitting it on the constrained devices the PR is aimed at.

The context is per-thread, so a lazy allocation inside `cn_slow_hash_v14` needs no locking.

src/wallet/wallet2.cpp
```diff
@@ -5051,6 +5115,13 @@ bool wallet2::finalize_multisig(const epee::wipeable_string &password, const std
 
 std::string wallet2::get_multisig_info() const
 {
+  // Nerva's multisig is the pre-rework legacy implementation (upstream
+  // replaced its key exchange after the 2022 disclosures and still flags
+  // even the fixed version experimental). New enrollment is disabled;
+  // existing multisig wallets keep working until HF14 so funds can move.
+  THROW_WALLET_EXCEPTION_IF(true, error::wallet_internal_error,
```
sn1f3rt, line 5122: **This throw escapes past its caller's error handling.**

`make_multisig` and `exchange_multisig_keys` are called from inside try/catch in simplewallet, and `wallet_rpc_server` gets one in this PR, and `api/wallet.cpp:1264` already had one. But `get_multisig_info()` is also called from `simple_wallet::prepare_multisig_main` (simplewallet.cpp:990), which does not catch, and neither does `simple_wallet::on_command`.

So the exception unwinds to epee's console catch-all at `contrib/epee/include/console_handler.h:420`, which only emits `LOG_ERROR("Exception at [console_handler], ...")`. The user running `prepare_multisig` gets a log line instead of the intended `fail_msg_writer()` message.

Wrapping the call in simplewallet, the way the sibling multisig commands already do, is enough. `gen_multisig` is fine, it has a top-level catch.

src/crypto/cna-vm.h
```diff
@@ -44,6 +44,32 @@
...
+// serial hops, then a program slice], each segment gated on a register the
+// program just mutated so the walk cannot advance without executing the
+// per-nonce random program (the GPU/ASIC barrier, same class as v6). All
+// values consensus-critical.
+#define CN_V7_HOPS           1024
+#define CN_V7_SEGMENTS       8
+// Per-nonce chase buffer: 24 MB. Big enough to spill any commodity L3 (so the
+// chase is DRAM-latency-bound, not cache/clock-bound) and to starve a GPU of
+// parallel nonces (VRAM / 24 MB), small enough to keep block verify ~0.3-0.5 s.
+#define CN_V7_BUFFER         (24 * 1024 * 1024)
```
sn1f3rt, line 80: **Flagging the verification cost as an explicit decision, because it runs against what HF13 was for.**

Built from this branch and measured locally:

```
v13 (CNA v6, 8 MB pad):    0.0473 s/hash   21.2 H/s
v14 (CNA v7, 24 MB chase): 0.1939 s/hash    5.2 H/s   -> 4.1x slower
```

That is on Apple Silicon, which has a favourable memory subsystem. An isolated model of just the chase, 3.1M dependent random accesses into 24 MB, came out around 0.40 s, which is closer to what a DRAM-bound x86 box should see.

PoW verification is single-threaded: one `m_hash_context` on `Blockchain`, used under the blockchain lock at `blockchain.cpp:3495` and `:1737`. There is no parallel batch PoW. At 60 s blocks that is 525,600 blocks per year, so roughly 7 hours of PoW per year of chain under v13 against roughly 28 hours under v14.

CNA v6 existed to fix sync, where `get_block_longhash` profiled at 45 percent of instructions and 99.68 percent of LLC read misses. The v7 chase is deliberately DRAM-bound, which is by construction an LLC-miss workload. `ASSUME_VALID_HEIGHT` plus quicksync cap the real exposure to chain since the last shipped checkpoint, so with a checkpoint per release it stays at a few hours, but it makes storing the PoW hash in `block_info` considerably more valuable than when it was filed as optional.

Not asking for a change here, just that this is an accepted trade rather than an unremarked one.

Related, for #116's gates rather than this PR: the 24 MB assumption is that it spills L3. AMD X3D desktop parts have 96 MB of L3 and server parts reach 256 to 384 MB, so four threads on an X3D would hold every buffer in cache and see cache latency instead of DRAM latency. The A57-versus-x86 measurement in the PR body does not cover that case. Worth one large-cache datapoint before a mainnet height is picked.

sn1f3rt (commented):
Reviewed against Monero master 52f02f2, fetched and diffed locally. The ported crypto checks out: `bulletproofs_plus.cc/.h`, `CLSAG_Gen`, `verRctCLSAGSimple`, `proveRctCLSAGSimple`, both triple-scalarmults and `device_default::clsag_*` are all byte-identical to upstream. Generator consistency is clean (BP+ commits with `rct::H`, the balance check uses the same point). The flag-day hardfork timing is correct: `HardFork::add` advances `current_fork_index` to `height + 1`, so a connecting block's transactions are checked against their own height's version, which is the same value `m_hardfork->check(bl)` pins `major_version` against. No lock-order inversion from the new `check_tx_outputs` call in `is_transaction_ready_to_go`, and `handle_alternative_block` runs the hardfork and checkpoint gates before `get_block_longhash`, so the more expensive PoW cannot be induced by a forged `major_version`.

Five items below: two that should land before merge, two small ones, and one design decision. Marked as a comment rather than a change request; the first two are one-liners.

Separately, and not inline: the PR body cites ported CLSAG and bulletproofs_plus test batteries, tamper and torsion cases, and cross-implementation interop. None of that is in the branch and the repo has no test tree, so it must have run out of tree. Diffing against upstream is stronger evidence than a test run for the ported code, but it says nothing about the Nerva-specific glue: the 6/7 renumbering through all three encoders, the fee and weight arithmetic, and CNA v7, which has no upstream to diff against. Could that harness be attached to #116 before a mainnet height is chosen, even as a scratch target outside the build?

src/crypto/slow-hash.c
mmokhi: Fixed. The v14 block now runs on a varied seed and OR-samples the buffer after the HW pass, so a fill regression that leaves it empty fails the test instead of passing vacuously. Re-ran on x86_64 and aarch64: HW and SW agree on the fully populated buffer. Good catch, this was the guard that mattered and it was guarding nothing.

src/crypto/slow-hash-impl.h
mmokhi: Fixed in both paths: the fill seed is now a fold of all 32 seed bytes, guarded so it can't start at the map's fixed point. The per-element loop is unchanged, so the fill cost and the checkpoint economics of the TMTO measurement don't move. Verified: seeds sharing the first 8 bytes but differing in the tail produce entirely different buffers now (0 of 4096 sampled qwords equal, was identical), and the all-zero seed fills non-degenerate. You're right on the timing argument too. It changes the hash, which is exactly why it lands now while no network has a height.

src/crypto/slow-hash.c
mmokhi: Done. The buffer allocates on first v14 use in the dispatcher (the context is per-thread, so no locking) and cn_hash_context_create no longer carries it, so the wallet-KDF contexts are back to ~9 MB. On allocation failure the hash aborts loudly instead of computing something wrong. The missing null-checks at the create sites predate this and are a separate cleanup.

src/wallet/wallet2.cpp
mmokhi: Done. prepare_multisig_main wraps the call the way make_multisig_main already does, so the refusal comes out through fail_msg_writer instead of unwinding to the console handler.

src/crypto/cna-vm.h
mmokhi: You're right that this should be an explicit decision rather than an unremarked one, so putting it on the record: the trade is accepted. #116's testnet gate includes recalibrating difficulty for the slower hash, and the verify cost is stated there. Agreed it makes storing the PoW hash in block_info considerably more attractive; that's now noted in #116 as a follow-up. On the large-L3 point: also agreed, a 24 MB buffer sits inside a 96 MB L3, so those parts chase at cache latency instead of DRAM. A large-cache datapoint is added to #116's gates before any mainnet height; if it matters in practice, buffer size is the knob, at the verify-cost price above.

mmokhi:
@sn1f3rt On the out-of-tree point: you're right, the test batteries ran outside the repo, so there is nothing in the branch to review them by. I'll attach the harness to #116 as a scratch target outside the build, so the Nerva-specific glue (the 6/7 renumbering, the fee arithmetic, CNA v7) has something reviewable. Agreed the upstream diff only vouches for the ported files.

The fixes above are pushed as fixup commits marking their home commits. Also re-measured after the seed change to confirm nothing moved: per-core gap on real hardware 1.34x (an aarch64 Cortex-A57 core at 1.48 GHz vs a 4.8 GHz x86-64 core), a 16 GB GPU at ~3.3x a 14-thread x86-64 box measured (~4x extrapolated at full VRAM, it is capacity-capped as designed), and the HW/SW self-test passes on the populated buffer on both architectures.

sn1f3rt (approved):

PR #118, opened 2026-07-22
chore: anchor assume-valid to the HF13 boundary

sn1f3rt:
HF13 (CryptoNight-Adaptive v6) activated at block **4,320,000** and the network is past it (`hard_fork_info`: `version: 13`, `enabled: true`, `earliest_height: 4320000`). This moves the assume-valid anchor up to the fork boundary so the entire pre-fork history skips PoW recomputation during sync.

## Changes
- `ASSUME_VALID_HEIGHT` `4290000` -> `4320000`
- Add checkpoint `4300000` = `6ee1e78bd980b5918bf2e7c87b72bb21bc86c7088750bea15e6b0b67a55a93c6` (100k cadence)
- Add checkpoint `4320000` = `45ae8c2a5f5463671ee092adc3f0f786c57369ba5a6fff2119bac0f1a2b86339` (HF13 activation)
- Drop the off-grid `4250000` / `4290000` checkpoints

## Why drop the off-grid entries
They were ad-hoc assume-valid anchors from previous releases. A checkpoint transitively pins every ancestor via `prev_id`, so the checkpoint at the fork boundary already anchors the whole skipped range — which is precisely why the startup guard only requires one checkpoint at or above `ASSUME_VALID_HEIGHT`. This restores the every-100k policy plus a single HF-boundary exception.

## Verification
- Both hashes cross-checked against two nodes.
- `4320000` is `major_version: 13`, `orphan_status: false` — the real fork block, not an orphan (there were 223 alt blocks around the fork).
- `4300000` is still `major_version: 12`, `orphan_status: false`.
- Startup guard holds: highest checkpoint (`4320000`) >= `ASSUME_VALID_HEIGHT` (`4320000`).
- Quicksync coupling from #115 is unaffected: `check_against_checkpoints` skips heights absent from the quicksync data and only fails on a hash conflict, so checkpoints above the quicksync range cannot break the loader.


mmokhi (approved):
Verified this end to end. Both hashes match what I get from two public nodes, 4300000 is v12 and 4320000 is v13, neither an orphan. The startup guard holds with the checkpoint and ASSUME_VALID_HEIGHT at the same height (it wants the max checkpoint at or above assume-valid, which this satisfies exactly), and the sync path skips strictly below the height, so the fork block itself still gets its PoW checked. That's the right place for the boundary, it's the first CNA v6 block. Dropping the off-grid entries is correct too: check_against_checkpoints skips heights it doesn't have and only fails on a conflict, so the transitive pinning argument is how the code actually behaves, and the reorg floor rises with the move.

PR #120, opened 2026-07-22
fix: force SSE2 float math on 32-bit x86 for difficulty determinism

### LINKED ISSUE #119: security: 32-bit x86 builds compute difficulty with x87 math and can fork from 64-bit nodes
## Summary

`next_difficulty_v6` — the active difficulty algorithm for HF6 and up (the chain is on HF13) — computes the next target in IEEE `double` and truncates it to `uint64_t` (`src/cryptonote_basic/difficulty.cpp:405-429`, dispatched at `blockchain.cpp` for `version >= 6`).

The 32-bit x86 toolchain (`i686-w64-mingw32 -m32`, `cmake/32-bit-toolchain.cmake`) does not pass `-msse2 -mfpmath=sse`, so 32-bit builds default to the x87 FPU and keep intermediate results in 80-bit extended precision. 64-bit builds use SSE2/NEON with 64-bit doubles.

When the computed `nextDifficulty` lands within ~1 ULP of an integer boundary, the x87 (32-bit) and SSE2 (64-bit) results can truncate to values that differ by 1. The two nodes then disagree on the difficulty target and on whether a block satisfies it (`check_hash`), and fork.

## Impact

A consensus chain split between 32-bit and 64-bit nodes. Per-block probability is low, but this is the live algorithm and 32-bit binaries are shipped (CI builds `i686 Win`, `i686 Linux`, `i686 musl`), so the exposure is real and ongoing.

## Affected

- `src/cryptonote_basic/difficulty.cpp` (`next_difficulty_v6`)
- `CMakeLists.txt` / `cmake/32-bit-toolchain.cmake` (no SSE2 float-math flags on 32-bit x86)

## Notes

Found in the internal security audit (item M1).

sn1f3rt:
Closes #119.

Forces `-msse2 -mfpmath=sse` on 32-bit x86 builds so their floating-point math matches the 64-bit SSE2 math used everywhere else, closing the difficulty-truncation divergence that can fork 32-bit vs 64-bit nodes.

`next_difficulty_v6` (the live algorithm for HF6+) computes the target in `double` and truncates to `uint64_t`. 32-bit x86 defaults to x87 (80-bit intermediate precision), so when the result lands within ~1 ULP of an integer boundary it can truncate one off from the 64-bit SSE2 result, and the two nodes disagree on `check_hash`.

## Why this and not an integer rewrite

Rewriting `next_difficulty_v6` to integer/fixed-point would change the difficulty values every node computes, i.e. a consensus change requiring a hard fork. Forcing SSE2 instead makes the divergent 32-bit builds agree with the 64-bit majority that already defines the canonical chain, so it deploys without a fork. `next_difficulty_v6` has no `a*b+c` pattern, so there is no FMA-contraction divergence left; the SSE2 flag is sufficient. (Integer math stays available as a belt-and-suspenders option for a future hard fork.)

## Scope

Applied only to 32-bit x86 (`NOT BUILD_64` and not ARM/PPC/S390X/RISCV/LOONGARCH). 64-bit x86 already uses SSE2 by default; ARM and others are unaffected (and `-msse2` would be invalid there).

## Testing

- `cmake` reconfigures clean; the block is correctly skipped on 64-bit and ARM.
- Gating verified across arches via a standalone check: 32-bit x86 -> `-msse2 -mfpmath=sse`; 64-bit x86 / 32-bit ARM / 64-bit ARM -> no change.
- The `i686 Win` / `i686 Linux` / `i686 musl` CI jobs exercise the actual 32-bit build with the new flags.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -794,6 +794,16 @@ else ()
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${ARCH_FLAG}")
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${ARCH_FLAG}")
 
+    # 32-bit x86 defaults to x87 (80-bit) float math, which can round differently
+    # from the 64-bit SSE2 math on other builds. next_difficulty_v6 truncates a
+    # double to the target, so that gap can fork 32-bit vs 64-bit nodes. Force SSE2
+    # so every x86 build computes difficulty identically.
+    if (NOT BUILD_64 AND NOT ARM AND NOT PPC64LE AND NOT PPC64 AND NOT PPC AND NOT S390X AND NOT RISCV AND NOT LOONGARCH)
+        message(STATUS "Forcing SSE2 float math on 32-bit x86 for cross-platform difficulty determinism")
+        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse2 -mfpmath=sse")
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse2 -mfpmath=sse")
+    endif ()
+
     set(WARNINGS "-Wall -Wextra -Wpointer-arith -Wundef -Wvla -Wwrite-strings -Wno-error=extra -Wno-error=deprecated-declarations -Wno-unused-parameter -Wno-error=unused-variable -Wno-error=undef -Wno-error=uninitialized")
     if (CMAKE_C_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         if (ARM)


mmokhi (approved):
Went through this carefully since it's consensus-adjacent. The #119 diagnosis checks out in the code: v6 accumulates 1/difficulty in doubles across the whole window and truncates at the end, exactly where 80-bit x87 intermediates can land one off from SSE2. The two claims that make the flag sufficient also hold: there's no a*b+c anywhere in the function so FMA contraction has nothing to grab (boost::math::round is floor/ceil based, exact), and the float math in difficulty.cpp is confined to v6, every older variant is integer, so the binary-wide flag closes the only consensus float site. The CMake block sits in the non-MSVC branch so the flags only reach gcc and clang, and I checked the CI logs for this head: "Forcing SSE2 float math" appears once in each of the three i686 jobs and not in the 64-bit ones, so the flags really land in the shipped builds. One conscious-choice note: this raises the 32-bit x86 floor to SSE2, anything much older than 2001 is out. The 32-bit machines we've actually seen all have SSE2 so I'm fine with it, but it belongs in the release notes. Matching the 64-bit majority instead of rewriting to integer math is the right call, the rewrite would itself be a consensus change.

PR #122, opened 2026-07-23
fix: harden restricted-mode RPC input validation

### LINKED ISSUE #121: security: restricted-mode RPC gaps - OOB read and missing request-size caps
## Summary

Several daemon RPC handlers (`src/rpc/core_rpc_server.cpp`) lack the input-size limits and bounds checks that keep a restricted (public) node from being used for memory-safety abuse or CPU/DB amplification. Found in the internal security audit (M3/M4/M5).

## M3 — `decode_outputs`: unbounded per-tx EC work on public nodes

`on_decode_outputs` has no restricted-mode gate or count cap. For every hash in `tx_hashes` (bounded only by the ~1 MB request body) it runs `generate_key_derivation` per tx pubkey, `derive_public_key` per output, and `decodeRct*` — thousands of scalar-mults per request, repeatable, i.e. a CPU-exhaustion amplifier on a public daemon. Separately, `boost::get<txout_to_key>(o.target)` is called outside the surrounding try/catch, so a non-standard output type throws `bad_get`. (Nerva-specific endpoint.)

## M4 — missing restricted-mode request-size caps

Unlike `get_outs` (which caps in restricted mode), these bulk handlers have no cap, so a public node can be forced into large DB scans / response builds (bounded only by the request body):

- `on_get_transactions` — no cap on `txs_hashes`.
- `on_get_blocks_by_height` — no cap on `heights`; reserves and fetches block + txs per height.
- `on_get_block_headers_range` — no range cap; the `restricted` local is computed but never used.

## M5 — `on_is_key_image_spent`: out-of-bounds read on a size mismatch

When a `key_images` entry parses as valid hex but is not 32 bytes, the handler sets an error status but does not `return`, then does `*reinterpret_cast<const crypto::key_image*>(b.data())` — reading 32 bytes past a shorter heap buffer. Restricted-reachable.

## Affected

- `src/rpc/core_rpc_server.cpp` — `on_decode_outputs`, `on_get_transactions`, `on_get_blocks_by_height`, `on_get_block_headers_range`, `on_is_key_image_spent`

## Notes

Found in the internal security audit (items M3, M4, M5).

sn1f3rt:
Closes #121.

Hardens daemon RPC input validation so a restricted (public) node can't be used for an out-of-bounds read or CPU/DB amplification.

## Changes (`src/rpc/core_rpc_server.cpp`)

**M5 — `is_key_image_spent` OOB read.** A `key_images` entry that parses as valid hex but is not 32 bytes set an error status without returning, then `reinterpret_cast`-ed 32 bytes past a shorter heap buffer. Added the missing `return` (matches the parse-failure path just above).

**M4 — restricted-mode request-size caps.** New constants `RESTRICTED_TRANSACTIONS_COUNT` (100), `RESTRICTED_BLOCK_COUNT` (1000), `RESTRICTED_BLOCK_HEADER_COUNT` (1000), matching upstream. Applied in the same `const bool restricted = m_restricted && ctx` style as `get_outs`:
- `get_transactions` — cap `txs_hashes`.
- `get_blocks_by_height` — cap `heights` (before the `reserve`).
- `get_block_headers_range` — cap the span, and hoist the previously-dead in-loop `restricted` local to where it's now used.

**M3 — `decode_outputs`.** Cap `tx_hashes` in restricted mode to bound the per-request EC work (`generate_key_derivation` / `derive_public_key` / `decodeRct*` per tx/output). Also replaced `boost::get<txout_to_key>(o.target).key` (throws `bad_get` on a non-standard output type) with the non-throwing pointer form, hoisted out of the inner derivation loop.

## Scope / notes

- `decode_outputs` is left **available but capped** on public nodes, not removed. It still takes a caller-supplied `sec_view_key` and decodes server-side, so it shouldn't be called on an untrusted node; a follow-up could gate it to unrestricted-only if desired.
- The adjacent Low items (L7 `relay_tx` length check, L8 `getblockhash` missing return) are intentionally not included here.

## Testing

- `obj_rpc` compiles clean.
- Input-validation changes verified by compile + inspection; each cap follows the proven `get_outs` restricted pattern and each handler keeps its own error convention (`return true` + `res.status` for the URI handlers, `return false` + `error_resp`/`res.status` for the JSON-RPC ones). End-to-end RPC exercise needs a live restricted daemon.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -63,8 +63,9 @@ using namespace epee;
 #define MAX_RESTRICTED_FAKE_OUTS_COUNT 40
 #define MAX_RESTRICTED_GLOBAL_FAKE_OUTS_COUNT 5000
 #define RESTRICTED_TRANSACTIONS_COUNT 100
+#define RESTRICTED_SPENT_KEY_IMAGES_COUNT 5000
 #define RESTRICTED_BLOCK_COUNT 1000
-#define RESTRICTED_BLOCK_HEADER_COUNT 1000
+#define RESTRICTED_BLOCK_HEADER_RANGE 1000
 
 #define OUTPUT_HISTOGRAM_RECENT_CUTOFF_RESTRICTION (3 * 86400) // 3 days max, the wallet requests 1.8 days
 
@@ -1119,6 +1120,11 @@ namespace cryptonote
       return ok;
 
     const bool restricted = m_restricted && ctx;
+    if (restricted && req.key_images.size() > RESTRICTED_SPENT_KEY_IMAGES_COUNT)
+    {
+      res.status = "Too many key images requested in restricted mode";
+      return true;
+    }
     const bool request_has_rpc_origin = ctx != NULL;
     std::vector<crypto::key_image> key_images;
     for(const auto& ki_hex_str: req.key_images)
@@ -2055,6 +2061,12 @@ namespace cryptonote
     };
 
     const bool restricted = m_restricted && ctx;
+    if (restricted && req.hashes.size() > RESTRICTED_BLOCK_COUNT)
+    {
+      error_resp.code = CORE_RPC_ERROR_CODE_WRONG_PARAM;
+      error_resp.message = "Too many block headers requested in restricted mode";
+      return false;
+    }
     if (!req.hash.empty())
     {
       if (!get(req.hash, res.block_header, restricted, error_resp))
@@ -2087,7 +2099,7 @@ namespace cryptonote
       return false;
     }
     const bool restricted = m_restricted && ctx;
-    if (restricted && req.end_height - req.start_height + 1 > RESTRICTED_BLOCK_HEADER_COUNT)
+    if (restricted && req.end_height - req.start_height > RESTRICTED_BLOCK_HEADER_RANGE)
     {
       error_resp.code = CORE_RPC_ERROR_CODE_TOO_BIG_HEIGHT;
       error_resp.message = "Too many block headers requested in restricted mode";
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -62,6 +62,9 @@ using namespace epee;
 
 #define MAX_RESTRICTED_FAKE_OUTS_COUNT 40
 #define MAX_RESTRICTED_GLOBAL_FAKE_OUTS_COUNT 5000
+#define RESTRICTED_TRANSACTIONS_COUNT 100
+#define RESTRICTED_BLOCK_COUNT 1000
+#define RESTRICTED_BLOCK_HEADER_COUNT 1000
 
 #define OUTPUT_HISTOGRAM_RECENT_CUTOFF_RESTRICTION (3 * 86400) // 3 days max, the wallet requests 1.8 days
 
@@ -514,6 +517,14 @@ namespace cryptonote
       return r;
 
     res.status = "Failed";
+
+    const bool restricted = m_restricted && ctx;
+    if (restricted && req.heights.size() > RESTRICTED_BLOCK_COUNT)
+    {
+      res.status = "Too many blocks requested in restricted mode";
+      return true;
+    }
+
     res.blocks.clear();
     res.blocks.reserve(req.heights.size());
     for (uint64_t height : req.heights)
@@ -653,6 +664,13 @@ namespace cryptonote
   {
     PERF_TIMER(on_decode_outputs);
 
+    const bool restricted = m_restricted && ctx;
+    if (restricted && req.tx_hashes.size() > RESTRICTED_TRANSACTIONS_COUNT)
+    {
+      res.status = "Too many transactions requested in restricted mode";
+      return false;
+    }
+
     std::vector<crypto::hash> vh;
     for(const auto& tx_hex_str: req.tx_hashes)
     {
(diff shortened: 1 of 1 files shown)


sn1f3rt:
Ran the PR's macOS-arm64 build (`nervad v0.3.0.0-70aa3663e`) with `--restricted-rpc --offline` and drove the endpoints directly. 4 of 5 verified end-to-end; M4b is the byte-identical cap pattern to the proven M4a.

| Finding | Test | Result |
|---|---|---|
| **M5** `is_key_image_spent` | short (4-hex) key image | "Failed, size of data mismatch", daemon stayed alive - OOB path gone |
| **M4a** `get_transactions` | 101 hashes / 1 hash | "Too many transactions requested in restricted mode" / OK |
| **M3** `decode_outputs` | 100 vs 101 hashes (valid address) | 100: status OK; 101: error - cap fires exactly at >100 |
| **M4c** `get_block_headers_range` | spans 999 / 1000 / 1001 / 2001 on a real 4.32M-block chain | 1000: OK (1000 headers); 1001: "Too many block headers requested in restricted mode" - boundary exact |
| **M4b** `get_blocks_by_height` | binary `.bin` endpoint | not driven directly; identical `if (restricted && req.heights.size() > RESTRICTED_BLOCK_COUNT)` pattern to M4a |

The M4a / M4c error strings are unique to this branch and restricted-only, so they also confirm the binary contains the change. The M3 and M4c boundary tests (passes at N, caps at N+1) isolate the cap from any downstream error.

mmokhi (commented):
Two caps missing that upstream has:

is_key_image_spent still takes req.key_images unbounded, in the handler this patches. Monero caps it at 5000, RESTRICTED_SPENT_KEY_IMAGES_COUNT. One db lookup per image plus a pool loop of pool_spent x requested, so the amplification stays open in the one function we touched.

get_block_header_by_hash takes req.hashes unbounded and reserves on its size, same shape you capped in get_blocks_by_height. Monero caps it at RESTRICTED_BLOCK_COUNT.

get_txids_loose is not in our tree, skip it.

Rest holds up. M5 is real, "aa" parses to 1 byte and the cast reads 32, and the return matches the path above. end - start + 1 cannot wrap, the validation above forces start <= end and both under bc_height. The ++i before the continue in decode_outputs is the bit that had to be right, i drives derive_public_key and decodeRctSimple.

Nits: end - start + 1 > 1000 against upstream end - start > 1000, ours allows 1000 headers and theirs 1001. Constant is RESTRICTED_BLOCK_HEADER_COUNT where upstream says RESTRICTED_BLOCK_HEADER_RANGE.

The OOB is worth running rather than reading. -DSANITIZE=ON and one is_key_image_spent with "aa", asan fires before and is quiet after.

src/rpc/core_rpc_server.cpp
```diff
@@ -62,6 +62,9 @@ using namespace epee;
 
 #define MAX_RESTRICTED_FAKE_OUTS_COUNT 40
 #define MAX_RESTRICTED_GLOBAL_FAKE_OUTS_COUNT 5000
+#define RESTRICTED_TRANSACTIONS_COUNT 100
+#define RESTRICTED_BLOCK_COUNT 1000
+#define RESTRICTED_BLOCK_HEADER_COUNT 1000
```
mmokhi: Upstream calls this RESTRICTED_BLOCK_HEADER_RANGE. Worth matching the name so future merges stay cheap.

src/rpc/core_rpc_server.cpp
```diff
@@ -2050,6 +2084,13 @@ namespace cryptonote
       error_resp.message = "Invalid start/end heights.";
       return false;
     }
+    const bool restricted = m_restricted && ctx;
+    if (restricted && req.end_height - req.start_height + 1 > RESTRICTED_BLOCK_HEADER_COUNT)
```
mmokhi: Upstream is end - start > RESTRICTED_BLOCK_HEADER_RANGE, no +1, so theirs allows 1001 headers and this allows 1000. Not wrong, just diverging from the thing we are porting.

mmokhi (changes requested):
Two nits on the header range, both cosmetic but easy while this is open.

src/rpc/core_rpc_server.cpp
sn1f3rt: Renamed to `RESTRICTED_BLOCK_HEADER_RANGE` in bc092f5.

src/rpc/core_rpc_server.cpp
sn1f3rt: Fixed in bc092f5 - now `end - start > RESTRICTED_BLOCK_HEADER_RANGE` (no +1), matching upstream (allows 1001).

sn1f3rt:
Addressed in bc092f5.

**Two more caps you flagged:**
- `is_key_image_spent`: capped `req.key_images` at `RESTRICTED_SPENT_KEY_IMAGES_COUNT` (5000) in restricted mode. Right that the OOB-read fix left the per-image lookup + `pool_spent x requested` loop open in the same handler.
- `get_block_header_by_hash`: capped `req.hashes` at `RESTRICTED_BLOCK_COUNT` (1000) before the `reserve` — same shape as `get_blocks_by_height`.

`get_txids_loose` isn't in our tree, skipped.

**Nits (both threads resolved):** renamed to `RESTRICTED_BLOCK_HEADER_RANGE`; dropped the `+1` so it's `end - start > RANGE`, matching upstream (allows 1001).

**ASAN, per your suggestion** - focused repro of the exact line (`*reinterpret_cast<const key_image*>(b.data())` on a short `blobdata`), built `-fsanitize=address`, input `"aa"`:

```
BEFORE (no return):
  ==ERROR: AddressSanitizer: stack-buffer-overflow ... READ of size 32
    #1 main m5_asan.cpp:23   (the reinterpret_cast)
    [96, 120) 'b' <== Memory access at offset 120 overflows this variable
AFTER (return added):
  size mismatch (1 != 32)      # returns; cast never reached, ASAN quiet
```

**Runtime verification** of the new/changed behavior (restricted `nervad`, real 4.32M-block chain for the range case):

| Handler | Test | Result |
|---|---|---|
| `is_key_image_spent` | 5001 / 5000 images | cap error / OK |
| `get_block_header_by_hash` | 1001 / 1000 hashes | cap error / passes cap (proceeds to lookup) |
| `get_block_headers_range` (post-`+1`-drop) | span 1001 / 1002 | OK (1001 headers) / cap error |

mmokhi (approved):
Looks good. Both caps are where they should be, key_images before the lookup loop and hashes before the reserve, and they each error the way their own handler does. Range check matches upstream now.

The sanitizer run is what really settles the out of bounds one, glad you did it.

Approving.

mmokhi (approved):

PR #124, opened 2026-08-12
Hard fork 14: ring size 16

### LINKED ISSUE #116: HF14: CLSAG + Bulletproofs+, and CryptoNight-Adaptive v7

mmokhi:
Ring size 16 for HF14, split out so it could be tested on its own. Part of #116.

The check in check_tx_inputs was unconditional, so it hit old blocks too. Raising the constant on its own would have invalidated every transaction already on chain and nothing could have synced. It goes by fork version now, 5 up to HF14 and 16 from HF14 on.

Only 16 is accepted from HF14, with no window where 5 still passes. HF14 already forces a wallet upgrade for CLSAG, so anything that can build a valid transaction at this version is a new wallet anyway, and there is no fork above 14 for a grace period to ever expire at. Monero enforces the exact size from the fork that introduces it.

Monero also allows a smaller ring for unmixable non-RCT outputs. We have none of those, so I left it out rather than add dead consensus code. That holds structurally rather than by assertion: check_tx_outputs rejects any output with a non-zero amount on every fork version, so there are no non-RCT outputs on the chain to strand and nobody's coins get stuck. It also means sweep_unmixable was already a dead path before this change, since the old rule required exactly 5 as well.

Wallets take the size from adjust_mixin so every spend path goes through one place instead of eleven callers remembering, and the fork version comes from get_current_hard_fork, same as the rest of the HF14 wallet code.

Cost: CLSAG goes from 224 to 576 bytes per input, a normal 2 in 2 out send from 1461 to 2209 bytes, so about 50% more fee. Being straight about it, most of our 4.4M rct outputs are coinbase, so the extra decoys are mostly mining outputs. Our real privacy limit is having 58,469 transactions in the whole chain, not the ring size.

This needs to ship before mainnet 4,400,000, which is roughly 39 days out, otherwise ring 16 has to move to its own fork version.

The testnet cannot test this as it stands. It carries ring 5 spends above the fork height, the first at 1018, so a node with this rule rejects that block and drops the peer. It needs relaunching first.


src/wallet/wallet2.cpp
```diff
@@ -10028,6 +10036,22 @@ bool wallet2::use_fork_rules(uint8_t version, int64_t early_blocks)
...
+// wallet code gets it. use_fork_rules would answer true on any version for a
+// light wallet, which would build the larger ring before consensus takes it.
+// One size is valid per version, so the same helper consensus uses decides it
+// here and the two cannot drift apart.
+uint64_t wallet2::adjust_mixin(uint64_t mixin)
+{
+  const uint64_t ring_size = cryptonote::get_ring_size(get_current_hard_fork());
+  if (mixin + 1 != ring_size)
+  {
+    MWARNING("Requested ring size " << (mixin + 1) << ", using " << ring_size);
```
sn1f3rt: `DEFAULT_MIXIN` stays 4 and every call site still passes it (`simplewallet.cpp:6069,6526,6775`, `wallet_rpc_server.cpp:1020,1073,1496,1551`, `api/wallet.cpp:1493,1497`), so from HF14 on every transfer and sweep hits `mixin + 1 != ring_size` and prints `Requested ring size 5, using 16` at WARNING level, which is on the console at the default log level.

That means a warning on every single send, for the entirely normal case. Suggest warning only when the caller asked for something other than the default, or dropping this to `MDEBUG`, so the message keeps meaning "you asked for X and did not get it".

src/wallet/wallet2.cpp
```diff
@@ -10028,6 +10036,22 @@ bool wallet2::use_fork_rules(uint8_t version, int64_t early_blocks)
...
 }
 //----------------------------------------------------------------------------------------------------
+// The fork version comes from the daemon the same way the rest of the HF14
+// wallet code gets it. use_fork_rules would answer true on any version for a
+// light wallet, which would build the larger ring before consensus takes it.
+// One size is valid per version, so the same helper consensus uses decides it
+// here and the two cannot drift apart.
+uint64_t wallet2::adjust_mixin(uint64_t mixin)
+{
+  const uint64_t ring_size = cryptonote::get_ring_size(get_current_hard_fork());
```
sn1f3rt: `get_current_hard_fork()` returns 0 when `m_offline` (wallet2.cpp:9979), and `get_ring_size(0)` is 5. So whenever the fork version is not available this silently builds the pre-fork ring and the daemon rejects the result with no hint why, which is the opposite of what the comment above promises.

The same fallback reaches `select_available_unmixable_outputs` and `discard_unmixable_outputs`, which then use a threshold of 5 instead of 16. Throwing on version 0 rather than defaulting to the old size would match the stated intent. `t_ring_size.cpp` never exercises `v = 0` either, since both loops start at 1.

contrib/hf14checks/t_ring_size.cpp
```diff
@@ -0,0 +1,137 @@
+// Ring size 16 at HF14: the consensus rule, and what it costs.
+//
+// Two things are pinned here:
+//  - the fork-gated rule in cryptonote_basic_impl, exhaustively over the fork
+//    versions and ring sizes that matter. The rule has to keep accepting the
+//    old size for historical blocks, accept either size at the fork that
+//    introduces the new one, and accept only the new one after that;
```
sn1f3rt: This says the rule has to "accept either size at the fork that introduces the new one", but `rule_across_fork_versions` asserts the opposite a few lines down: `CHECK_FALSE(is_valid_ring_size(HF_VERSION_RING_SIZE_16, DEFAULT_RINGSIZE))`. The PR description is explicit that there is deliberately no such window.

Stale comment describing the design that was rejected. Worth fixing given it documents a consensus rule.

contrib/hf14checks/t_ring_size.cpp
```diff
@@ -0,0 +1,137 @@
...
+    verRctCLSAGSimple(msg_new, new_sig, pubs_new, Cout_new);
+    auto c = std::chrono::steady_clock::now();
+    t_old.push_back(std::chrono::duration<double, std::milli>(b - a).count());
+    t_new.push_back(std::chrono::duration<double, std::milli>(c - b).count());
+  }
+  const double m_old = median(t_old), m_new = median(t_new);
+
+  std::printf("   ring %2d: %4zu bytes/input, verify %.3f ms\n", DEFAULT_RINGSIZE, old_bytes, m_old);
+  std::printf("   ring %2d: %4zu bytes/input, verify %.3f ms\n", DEFAULT_RINGSIZE_V14, new_bytes, m_new);
+  std::printf("   delta  : +%zu bytes/input, verify %.2fx\n", new_bytes - old_bytes, m_new / m_old);
```
sn1f3rt, line 139: The file header says the verification cost is pinned "so a regression in either is visible", but nothing here checks it: the medians are computed, printed, and dropped. The only assertions are on the byte sizes, and those are tautological given the CLSAG layout, since `s.size()` is the ring size by construction. A 10x verification regression passes ctest silently.

Either assert a bound on `m_new / m_old` (the expected ratio leaves plenty of headroom for a loose one) or drop the claim from the header comment.

Minor: `std::sort` is used here without `<algorithm>`; it only compiles on transitive includes.

src/wallet/wallet2.cpp
```diff
@@ -10141,14 +10165,14 @@ const wallet2::transfer_details &wallet2::get_transfer_details(size_t idx) const
 //----------------------------------------------------------------------------------------------------
 std::vector<size_t> wallet2::select_available_unmixable_outputs()
 {
-  // request all outputs with less than 3 instances
-  return select_available_outputs_from_histogram(DEFAULT_MIXIN + 1, false, true, false);
+  // too few of the amount around to build a ring of the size this fork wants
+  return select_available_outputs_from_histogram(cryptonote::get_ring_size(get_current_hard_fork()), false, true, false);
```
sn1f3rt, line 10177: With this threshold at 16, `sweep_unmixable` is now hunting for non-RCT amounts with fewer than 16 instances, a set that is provably empty: `check_tx_outputs` rejects `o.amount != 0` unconditionally on every fork version (blockchain.cpp:2784), so the chain has no non-RCT outputs at all. (That same rule is what makes the "nobody's coins get stuck" claim in the description hold structurally rather than by assertion, which is worth stating somewhere.)

Meanwhile `create_unmixable_sweep_transactions` passes `0 /*fake_outs_count*/` into `create_transactions_from`, which `adjust_mixin` now rewrites to 15, while the command help still reads "Send all unmixable outputs to yourself with ring_size 1" (simplewallet.cpp:2860).

The path was already dead, since the old check required exactly 5 too, so this is cosmetic. But if it is being touched, removing `sweep_unmixable` reads better than leaving it pointed at a rule it cannot satisfy.

src/wallet/wallet2.cpp
mmokhi: Every send carries DEFAULT_MIXIN so it fired on all of them. Warns only when the caller picked a size itself now.

src/wallet/wallet2.cpp
mmokhi: Offline it got 0 and built ring 5, then the daemon dropped it. Throws no_connection_to_daemon now.

Cold signing unaffected, adjust_mixin is only reached from create_transactions_2 and create_transactions_from, sign_tx and load_unsigned_tx never call it.

discard_unmixable_outputs is gone with the sweep command. select_available_unmixable_outputs still reads 5 offline, its first act is a histogram rpc so it fails there before the threshold is used.

Tests cover v=0 now, they started at 1.

contrib/hf14checks/t_ring_size.cpp
mmokhi: Fixed, it was left from the grace period version.

contrib/hf14checks/t_ring_size.cpp
mmokhi: Asserts m_new/m_old < 8 now, real is about 3.2 so it won't flake. Added <algorithm>.

src/wallet/wallet2.cpp
mmokhi: Removed the command. discard_unmixable_outputs went with it, nothing else called it, and select_available_mixable_outputs had no callers at all.

sweep_dust rpc and createSweepUnmixableTransaction stay, both still reach create_unmixable_sweep_transactions.

The amount != 0 reasoning is in the description now, it's why the path was already dead, the old rule needed exactly 5 too.

src/wallet/wallet2.cpp
```diff
@@ -10180,15 +10206,6 @@ std::vector<wallet2::pending_tx> wallet2::create_unmixable_sweep_transactions()
   return create_transactions_from(m_account_public_address, false, 1, unmixable_transfer_outputs, unmixable_dust_outputs, 0 /*fake_outs_count */, 0 /* unlock_time */, 1 /*priority */, std::vector<uint8_t>());
```
sn1f3rt, line 10206: The CLI command is gone, but this builder is still reachable from the `sweep_unmixable` JSON-RPC method (`wallet_rpc_server.h:97` -> `on_sweep_dust` -> `wallet_rpc_server.cpp:1449`) and from `createSweepUnmixableTransaction` (`api/wallet.cpp:1584`). Both land here, and the `0` still becomes 15 in `adjust_mixin` — so an RPC named `sweep_unmixable` now builds a full-ring transaction, which is the one thing an unmixable sweep cannot be. It warns, then fails with `not_enough_outs`, drawing from a set that is provably empty anyway.

Keeping the API surface for compatibility is reasonable, but the result is backwards from the usual compat argument: the CLI can no longer reach the path and the RPC still can. If the goal is not breaking callers, having `on_sweep_dust` return an empty result (or an explicit "no unmixable outputs exist on this chain") is cheaper than routing through a builder that constructs a ring-16 tx and fails downstream.

Minor, two lines below: removing `discard_unmixable_outputs` left a double blank line and dropped the `//----` separator before `get_tx_key_cached`.

src/simplewallet/simplewallet.cpp
```diff
@@ -552,7 +552,6 @@ void simple_wallet::handle_transfer_exception(const std::exception_ptr &e, bool
       {
         writer << "\n" << sw::tr("output amount") << " = " << print_money(outs_for_amount.first) << ", " << sw::tr("found outputs to use") << " = " << outs_for_amount.second;
       }
```
sn1f3rt, line 554: Correct to drop the pointer to a command that no longer exists, but `api/wallet.cpp:1546` still appends `"Please sweep unmixable outputs."` to the same `not_enough_outs_to_mix` message. Worth removing too, so the two paths don't give contradictory advice.

Also `contrib/rlwrap/monerocommands_monero-wallet-cli.txt:29` still lists `sweep_unmixable` in the completion word list.

src/simplewallet/simplewallet.cpp
```diff
@@ -2855,9 +2854,6 @@ simple_wallet::simple_wallet()
   m_cmd_binder.set_handler("transfer", std::bind(&simple_wallet::on_command, this, &simple_wallet::transfer, std::placeholders::_1),
                            tr(USAGE_TRANSFER),
                            tr("Transfer <amount> to <address>. If the parameter \"index=<N1>[,<N2>,...]\" is specified, the wallet uses outputs received by addresses of those indices. If omitted, the wallet randomly chooses address indices to be used. In any case, it tries its best not to combine outputs across multiple addresses. <priority> is the priority of the transaction. The higher the priority, the higher the transaction fee. Valid values in priority order (from lowest to highest) are: default, low, medium, high. If omitted, the default value (see the command \"set priority\") is used. <ring_size> is the number of inputs to include for untraceability. Multiple payments can be made at once by adding URI_2 or <address_2> <amount_2> etcetera (before the payment ID, if it's included)"));
```
sn1f3rt: This still explains `<ring_size> is the number of inputs to include for untraceability`, but `USAGE_TRANSFER` (line 169) doesn't accept the argument and `transfer` never parses one. Same for `USAGE_SWEEP_BELOW`, `USAGE_SWEEP_SINGLE` and `USAGE_DONATE` (171-173), which all advertise `[<ring_size>]` while `sweep_main` (6296-6320) and `sweep_single` (6550-6560) read only `index=`, priority and `outputs=`.

Pre-existing, but this PR makes it actively wrong: ring size is now fixed by consensus and the `adjust_mixin` warning is deliberately silent for the default, so a user who believes the help text and types a size gets no feedback at all. Worth deleting the `[<ring_size>]` tokens and that help sentence while you're here.

src/wallet/wallet2.cpp
mmokhi: on_sweep_dust answers empty now instead of going through the builder. Blank line and separator fixed.

Left createSweepUnmixableTransaction on the builder. Doing the same there orphans about 60 catch blocks for exceptions that can no longer fire, and the result is the same either way.

src/simplewallet/simplewallet.cpp
mmokhi: Both gone, api/wallet.cpp and the rlwrap list.

src/simplewallet/simplewallet.cpp
mmokhi: Dropped the tokens from sweep_below, sweep_single and donate, and the sentence from transfer's help. None of them parse a size, they all pass DEFAULT_MIXIN.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -1446,7 +1446,11 @@ namespace tools
 
     try
     {
-      std::vector<wallet2::pending_tx> ptx_vector = m_wallet->create_unmixable_sweep_transactions();
+      // An unmixable output is one there are too few of its amount around to
+      // build a ring from. check_tx_outputs refuses a non-zero output amount on
+      // every fork version, so this chain has none. Answer empty rather than
+      // walk a builder that would go on to ask for a full ring.
```
sn1f3rt: The short-circuit is right, but the last sentence describes a control flow that does not happen. `create_unmixable_sweep_transactions` early-returns an empty vector when `select_available_unmixable_outputs()` comes back empty (`wallet2.cpp:10190-10193`), which on this chain is always, so it never reaches `create_transactions_from` and never asks for a full ring.

My earlier comment said this path "warns, then fails with `not_enough_outs`", which was wrong and is what this sentence is echoing. What the change actually buys is skipping a pointless `get_output_histogram` round-trip and stating the intent at the call site, both worth having. Suggest just dropping "rather than walk a builder that would go on to ask for a full ring" so the comment doesn't outlive the mistake.

Not blocking.

sn1f3rt (approved):

src/wallet/wallet_rpc_server.cpp
mmokhi: Dropped it. I had the early return checked too, so my comment repeated the same wrong bit.

It saves the get_output_histogram round trip, that is what it says now. In 1dc710f.

sn1f3rt:
Pushed to this branch while you are away, since we want to land it.

Rebased onto master (40c2e91) to pick up #134, #133, #136 and #132. Clean, no conflicts, and the HF14 mainnet height in the result is 4500000 from #134 rather than the 4400000 this branch was written against.

Added one commit, `fix(wallet): drop the last pointer to an unmixable sweep`. The `not_enough_outs_to_mix` handler in wallet_rpc_server still appended "Please use sweep_dust." — the third copy of the advice removed from simplewallet and api/wallet, missed because it names the rpc method rather than the cli command. It now sends a caller to a method that answers empty.

Verified locally on the rebased tree: full Release build of nervad, nerva-wallet-cli, nerva-wallet-rpc and libwallet_api with no errors and no new warnings in the touched files, and t_ring_size passes 312 checks, measuring 224 to 576 bytes per input and 3.15x verify, which matches the 3.17x in the commit message.

sn1f3rt (approved):
Re-approving on d8e0bc1. The earlier approval was against 660a586, before 1dc710f and before the rebase.

Reviewed the current tree independently rather than re-reading the diff. The fork boundary lands exactly where it should: HardFork::add(block, H) sets current_fork_index from get_voted_fork_index(H+1) after the block is stored, so get_current_version() during validation of block H is version(H). The first HF14 block requires 16 and the last HF13 block requires 5, with no overlap. hard_fork_info hands the wallet that same value, and both sides size the ring through get_ring_size, so the two cannot disagree at the boundary.

Historical resync is unchanged, since below the fork the rule is what the old unconditional check already enforced, and the same gating pattern has been live across HF13 at 4320000. Pooled ring-5 transactions cannot survive the crossing either: m_input_cache clears on every block, is_transaction_ready_to_go re-runs the input check, and validate() fires on the version change from #103.

Leaving out the unmixable exemption holds structurally, not by assertion, since check_tx_outputs_except_subgroup refuses a non-zero output amount on every fork version. The new loop is also more robust than the one it replaces, which called boost::get on every input and would have thrown on a txin_gen.

Built the rebased tree locally, no errors and no new warnings in the touched files, and t_ring_size passes 312 checks at 224 to 576 bytes per input and 3.15x verify.

Note for whoever schedules the fork: the testnet still cannot exercise this, it carries ring 5 spends above height 1000. That is not new, master already moved 12 and 13 to 700 and 800 against the retired chain, so a fresh genesis was required regardless.

PR #126, opened 2026-08-14
Coinbase sum over a long range overflows 64 bits

mmokhi:
`get_coinbase_tx_sum` adds up coinbase minus fees over a block range. Asked for the whole chain it answered 796076.392758448384 XNV against a real 19242820.466468, short by exactly 2^64 atomic units, because the accumulators were `uint64` and the emission passed that a while ago.

Summing the same range in chunks of 100000 gives the right total, since no single chunk overflows. That is what the supply tooling does, which is why its number has been right while the daemon's has not.

Accumulators and return type are 128 bit now. Per block amounts stay `uint64`, and `MONEY_SUPPLY` and `already_generated_coins` are untouched, so none of this reaches consensus and it can ship in a point release.

### Over RPC

`emission_amount` and `fee_amount` still carry the low 64 bits, so anything reading just those sees no change. The high half arrives in the new `emission_amount_top64` and `fee_amount_top64` — the same `<name>` + `<name>_top64` shape this tree already uses for `cumulative_difficulty`, rather than Monero's `wide_*` hex string, which this tree does not use on any wire. `store_difficulty` is generalised to `store_128` and now serves both. `CORE_RPC_VERSION_MINOR` goes to 1 for the added fields.

The daemon CLI rebuilds the value from the two halves rather than trusting a single wide field, so it still prints the right number against a daemon that predates this change, where the high halves are absent and default to 0.

### Measured

On a copy of mainnet at height 4343869, before and after:

    emission_amount_top64      0            ->  1
    full range total           796076.39    ->  19242820.466468 XNV
    difference                              18446744073709551616, exactly 2^64
    low 64 bits                unchanged
    one call vs 44 chunks      disagreed    ->  equal

The 19242820.466468 matches what the supply tooling derives from its own checkpoint at block 4300000, which was never fed into the code.

Verified locally: the four changed translation units plus `wallet2`, `simplewallet`, `wallet_rpc_server`, `blockchain`, `blockchain_import` and `rctSigs` compile clean; the `uint64` `print_money` path is byte-identical to before across the range; the 128 bit split round-trips at 0, 1, 2^63, 2^64 ± 1 and 2^128 - 1; and an old-daemon response parsed through the real epee serializer leaves the new fields at 0 and reassembles to exactly the value it sent.

### Left out on purpose

Our bounds check rejects only `count > chain height`, where Monero rejects `count > chain height - height`. Ours accepts a range running past the tip and answers OK with a short sum. Worth fixing, but it turns a call that succeeds today into an error, so it does not belong in the same change.

`emission_amount += coinbase_amount - tx_fee_amount` is still `uint64` arithmetic. With a `uint64` accumulator an underflow there cancelled out mod 2^64; with a 128 bit one it would not. Safe today because `validate_miner_transaction` rejects any block where `money_in_use != base_reward + fee`, and this tree never sets `partial_block_reward` — noted so that check is not relaxed later without revisiting the line.

### DIFF
--- a/src/cryptonote_core/cryptonote_core.cpp
+++ b/src/cryptonote_core/cryptonote_core.cpp
@@ -1189,10 +1189,12 @@ namespace cryptonote
     return m_mempool.check_for_key_images(key_im, spent);
   }
   //-----------------------------------------------------------------------------------------------
-  std::pair<uint64_t, uint64_t> core::get_coinbase_tx_sum(const uint64_t start_offset, const size_t count)
+  std::pair<boost::multiprecision::uint128_t, boost::multiprecision::uint128_t> core::get_coinbase_tx_sum(const uint64_t start_offset, const size_t count)
   {
-    uint64_t emission_amount = 0;
-    uint64_t total_fee_amount = 0;
+    // A sum over the whole chain passed what 64 bits hold, so the running
+    // totals have to be wider than the per block amounts they add up.
+    boost::multiprecision::uint128_t emission_amount = 0;
+    boost::multiprecision::uint128_t total_fee_amount = 0;
     if (count)
     {
       const uint64_t end = start_offset + count - 1;
@@ -1214,7 +1216,7 @@ namespace cryptonote
       });
     }
 
-    return std::pair<uint64_t, uint64_t>(emission_amount, total_fee_amount);
+    return std::pair<boost::multiprecision::uint128_t, boost::multiprecision::uint128_t>(emission_amount, total_fee_amount);
   }
 
   uint64_t core::get_generated_coins()
--- a/src/daemon/rpc_command_executor.cpp
+++ b/src/daemon/rpc_command_executor.cpp
@@ -40,6 +40,7 @@
 #include "cryptonote_basic/difficulty.h"
 #include "cryptonote_basic/hardfork.h"
 #include <boost/format.hpp>
+#include <boost/multiprecision/cpp_int.hpp>
 #include <ctime>
 #include <string>
 
@@ -1857,11 +1858,16 @@ bool t_rpc_command_executor::print_coinbase_tx_sum(uint64_t height, uint64_t cou
     }
   }
 
+  // A daemon from before the sum was widened leaves the top halves at 0, which
+  // reassembles to exactly the value it sent.
+  const boost::multiprecision::uint128_t emission = (boost::multiprecision::uint128_t(res.emission_amount_top64) << 64) | res.emission_amount;
+  const boost::multiprecision::uint128_t fees = (boost::multiprecision::uint128_t(res.fee_amount_top64) << 64) | res.fee_amount;
+
   tools::msg_writer() << "Sum of coinbase transactions between block heights ["
     << height << ", " << (height + count) << ") is "
-    << cryptonote::print_money(res.emission_amount + res.fee_amount) << " "
-    << "consisting of " << cryptonote::print_money(res.emission_amount) 
-    << " in emissions, and " << cryptonote::print_money(res.fee_amount) << " in fees";
+    << cryptonote::print_money(emission + fees) << " "
+    << "consisting of " << cryptonote::print_money(emission)
+    << " in emissions, and " << cryptonote::print_money(fees) << " in fees";
   return true;
 }
 
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -87,7 +87,7 @@ namespace cryptonote
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
 #define CORE_RPC_VERSION_MAJOR 4
-#define CORE_RPC_VERSION_MINOR 0
+#define CORE_RPC_VERSION_MINOR 1
 #define MAKE_CORE_RPC_VERSION(major,minor) (((major)<<16)|(minor))
 #define CORE_RPC_VERSION MAKE_CORE_RPC_VERSION(CORE_RPC_VERSION_MAJOR, CORE_RPC_VERSION_MINOR)
 
@@ -1957,12 +1957,16 @@ namespace cryptonote
     struct response_t: public rpc_response_base
     {
       uint64_t emission_amount;
+      uint64_t emission_amount_top64;
       uint64_t fee_amount;
+      uint64_t fee_amount_top64;
 
       BEGIN_KV_SERIALIZE_MAP()
         KV_SERIALIZE_PARENT(rpc_response_base)
         KV_SERIALIZE(emission_amount)
+        KV_SERIALIZE(emission_amount_top64)
         KV_SERIALIZE(fee_amount)
+        KV_SERIALIZE(fee_amount_top64)
       END_KV_SERIALIZE_MAP()
     };
     typedef epee::misc_utils::struct_init<response_t> response;
--- a/src/cryptonote_basic/cryptonote_format_utils.cpp
+++ b/src/cryptonote_basic/cryptonote_format_utils.cpp
@@ -1029,17 +1029,29 @@ namespace cryptonote
     }
   }
   //---------------------------------------------------------------
-  std::string print_money(uint64_t amount, unsigned int decimal_point)
+  static void insert_money_decimal_point(std::string &s, unsigned int decimal_point)
   {
     if (decimal_point == (unsigned int)-1)
       decimal_point = default_decimal_point;
-    std::string s = std::to_string(amount);
     if(s.size() < decimal_point+1)
     {
       s.insert(0, decimal_point+1 - s.size(), '0');
     }
     if (decimal_point > 0)
       s.insert(s.size() - decimal_point, ".");
+  }
+  //---------------------------------------------------------------
+  std::string print_money(uint64_t amount, unsigned int decimal_point)
+  {
+    std::string s = std::to_string(amount);
+    insert_money_decimal_point(s, decimal_point);
+    return s;
+  }
+  //---------------------------------------------------------------
+  std::string print_money(const boost::multiprecision::uint128_t &amount, unsigned int decimal_point)
+  {
+    std::string s = amount.str();
+    insert_money_decimal_point(s, decimal_point);
     return s;
   }
   //---------------------------------------------------------------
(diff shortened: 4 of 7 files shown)


src/rpc/core_rpc_server_commands_defs.h
```diff
@@ -1957,12 +1957,20 @@ namespace cryptonote
     struct response_t: public rpc_response_base
     {
       uint64_t emission_amount;
+      std::string wide_emission_amount;
```
sn1f3rt: This adds four fields to a response struct, so `CORE_RPC_VERSION_MINOR` needs to go 1 -> 2 (line 90 of this file). The policy comment sitting directly above it says to bump minor on *any* change here, and #83 did exactly that for its response change.

It matters concretely rather than just as bookkeeping: without a bump a client has no way to tell whether the daemon it is talking to returns the wide fields, which is the same gap behind the `rpc_command_executor.cpp` comment below.

src/daemon/rpc_command_executor.cpp
```diff
@@ -1859,9 +1859,9 @@ bool t_rpc_command_executor::print_coinbase_tx_sum(uint64_t height, uint64_t cou
 
   tools::msg_writer() << "Sum of coinbase transactions between block heights ["
     << height << ", " << (height + count) << ") is "
-    << cryptonote::print_money(res.emission_amount + res.fee_amount) << " "
-    << "consisting of " << cryptonote::print_money(res.emission_amount) 
-    << " in emissions, and " << cryptonote::print_money(res.fee_amount) << " in fees";
+    << cryptonote::print_money(boost::multiprecision::uint128_t(res.wide_emission_amount) + boost::multiprecision::uint128_t(res.wide_fee_amount)) << " "
+    << "consisting of " << cryptonote::print_money(boost::multiprecision::uint128_t(res.wide_emission_amount))
+    << " in emissions, and " << cryptonote::print_money(boost::multiprecision::uint128_t(res.wide_fee_amount)) << " in fees";
```
sn1f3rt: In the `m_is_rpc` path `res` comes from whatever daemon is on the other end, and reading only `wide_*` breaks against one that predates this change. `KV_SERIALIZE_N` in epee discards the serializer's return value, so a field the peer does not send is silently left at its default, and `boost::multiprecision::uint128_t("")` parses as 0 rather than throwing (checked against boost 1.90). A new CLI pointed at an old daemon therefore prints `0.000000000000` where it used to print the right number for any range under 2^64 - a wrong answer rather than an error.

Both halves are already in the response, so this can be reconstructed without parsing the string at all. That is correct against old daemons too, since `top64` defaults to 0 and `emission_amount` is exactly what they send, and it drops the `std::runtime_error` a malformed `wide_*` from a remote daemon currently throws out of this function.

```suggestion
  const boost::multiprecision::uint128_t emission = (boost::multiprecision::uint128_t(res.emission_amount_top64) << 64) | res.emission_amount;
  const boost::multiprecision::uint128_t fees = (boost::multiprecision::uint128_t(res.fee_amount_top64) << 64) | res.fee_amount;

  tools::msg_writer() << "Sum of coinbase transactions between block heights ["
    << height << ", " << (height + count) << ") is "
    << cryptonote::print_money(emission + fees) << " "
    << "consisting of " << cryptonote::print_money(emission)
    << " in emissions, and " << cryptonote::print_money(fees) << " in fees";
```

Compiled that against boost 1.90 to confirm it builds and reassembles the value correctly.

src/cryptonote_core/cryptonote_core.cpp
```diff
@@ -1189,10 +1189,12 @@ namespace cryptonote
...
   //-----------------------------------------------------------------------------------------------
-  std::pair<uint64_t, uint64_t> core::get_coinbase_tx_sum(const uint64_t start_offset, const size_t count)
+  std::pair<boost::multiprecision::uint128_t, boost::multiprecision::uint128_t> core::get_coinbase_tx_sum(const uint64_t start_offset, const size_t count)
   {
-    uint64_t emission_amount = 0;
-    uint64_t total_fee_amount = 0;
+    // A sum over the whole chain passed what 64 bits hold, so the running
+    // totals have to be wider than the per block amounts they add up.
+    boost::multiprecision::uint128_t emission_amount = 0;
+    boost::multiprecision::uint128_t total_fee_amount = 0;
```
sn1f3rt, line 1197: Non-blocking, but worth recording since the widening shifts it: `emission_amount += coinbase_amount - tx_fee_amount` further down is still uint64 arithmetic. With a uint64 accumulator an underflow there cancelled out mod 2^64; with a 128 bit one it does not, and a single such block would add roughly 1.8e19 to the total.

Safe today because `validate_miner_transaction` rejects any block where `money_in_use != base_reward + fee` (`src/cryptonote_core/blockchain.cpp:1260`) - upstream permits partial reward claims, we do not. Flagging it only so that check does not get relaxed later without revisiting this line.

sn1f3rt (changes requested):
The fix is correct and reads as a faithful backport of upstream's `store_128` handling. Two things I would like addressed before merge, plus a note and a follow-up.

**What I verified**

- Compiled the three changed translation units plus `wallet2.cpp` and `simplewallet.cpp` against this branch using the project's own flags from `compile_commands.json`. All clean, and there is no `print_money` overload ambiguity at any call site in the tree: every argument is of `uint64_t` rank or better, so the existing overload always wins over the user-defined conversion to `uint128_t`.
- `cryptonote::hex()` here is `hex(difficulty_type_128)` (`src/cryptonote_basic/difficulty.h:43`), the same type, emitting `0x`-prefixed lowercase. It round-trips through boost 1.90's `uint128_t` string constructor, radix prefix included.
- The low 64 bits written by `store_128` are bit-identical to the value `emission_amount` carried before, so existing readers genuinely see no change.

**On the consensus claim**

It holds, and for a reason worth stating explicitly: `compute_generated_coins` caps `already_generated_coins` at `MONEY_SUPPLY` (`src/cryptonote_core/blockchain.cpp:3741`), so that value never wrapped. That also makes the numbers in the description internally consistent, since 19242820.466468 minus the 2^64 cap leaves exactly the 796076.392758 of post-cap tail emission the daemon has been reporting.

**Follow-up, not for this PR**

Agreed on leaving the `req.count > bc_height` bounds check alone, but it deserves its own issue now rather than later. A range running past the tip quietly returning a short sum is more dangerous once the number is trustworthy enough that people act on it.

**Nit**

`cryptonote_format_utils.h` now pulls `boost/multiprecision/cpp_int.hpp` into one of the most widely included headers in the tree. Probably acceptable given `difficulty.h` already does something similar, but it is a real build-time cost.

src/rpc/core_rpc_server_commands_defs.h
sn1f3rt: Done, but the target moved: #132 bumped the major to 4 and reset minor to 0, so the branch was rebased first and the bump is 0 -> 1, not 1 -> 2. Also relevant here, the four fields became two: `wide_emission_amount` / `wide_fee_amount` are gone and only `emission_amount_top64` / `fee_amount_top64` were added, matching the `<name>` + `<name>_top64` shape this tree already uses for `cumulative_difficulty`. Nothing in the tree emits a `wide_*` field on any wire.

src/daemon/rpc_command_executor.cpp
sn1f3rt: Taken, and the underlying hazard is gone rather than worked around: the `wide_*` fields were dropped entirely, so there is no hex string left to mis-parse and no `std::runtime_error` path out of this function. The executor now reassembles from `(uint128_t(top64) << 64) | low`, which is the idiom already used at five sites in this file for `cumulative_difficulty`.

Verified end to end rather than by inspection: an old-daemon response parsed through the real epee serializer into the real response struct leaves both `top64` fields at 0 and reassembles to exactly the value the old daemon sent.

src/cryptonote_core/cryptonote_core.cpp
sn1f3rt: Left the code as it stands, and the note holds. Traced it to be sure it is unreachable rather than merely unlikely: `coinbase_amount` is `get_outs_money_amount(miner_tx)` = `money_in_use`, `tx_fee_amount` is the block fee sum = `fee`, and `validate_miner_transaction` rejects `base_reward + fee != money_in_use`, so the subtraction is `base_reward` and cannot go negative. The check is skipped only at genesis, which has no non-coinbase txs, and a missed tx in `get_transactions` shrinks the fee sum rather than growing it.

So it cannot fire on any chain that passed validation. Recording it here rather than opening an issue, since there is no present defect to describe - it is a constraint on relaxing that check later.

sn1f3rt (approved):
Both blocking points are addressed, and the shape of the fix changed on the way there.

`wide_emission_amount` / `wide_fee_amount` are gone. This tree exposes no `wide_*` field on any wire: `store_difficulty` never emitted one, and the `wide_difficulty` members in `message_data_structs.h` are internal `difficulty_type_128` working fields that `json_object.cpp` does not serialize. The response now carries `emission_amount_top64` / `fee_amount_top64` only, matching the `<name>` + `<name>_top64` shape already used for `cumulative_difficulty`, and `store_difficulty` is generalised to `store_128` to serve both. That also removes the hex-string parse the executor comment was about.

The version bump is 0 -> 1, not the 1 -> 2 I asked for: #132 moved the major to 4 and reset the minor, so the branch was rebased onto master first.

What I checked on the final head:

- All 19 CI jobs green, including `i686 Linux` / `i686 Win` / `i686 musl` and both ARM v7 targets, which is the real gate for the 128-bit arithmetic on 32-bit platforms.
- The `uint64` `print_money` path is byte-identical to before across 9 values and 5 decimal-point settings; the refactor into `insert_money_decimal_point` changes nothing.
- The 128-bit split round-trips at 0, 1, 2^63, 2^64 - 1, 2^64, 2^64 + 1 and 2^128 - 1.
- Old-daemon compatibility end to end through the real epee serializer and the real response struct, not by inspection: absent `top64` fields land at 0 and the executor reassembles to exactly the value the old daemon sent.
- Overload resolution and linkage: `cryptonote_format_utils.o` defines the `uint128_t` overload and `rpc_command_executor.o` references the identical mangled symbol, so `emission + fees` binds to the wide overload rather than silently narrowing.

One correction to the record: the exact wrapped figure is 796076.392758448384, not the 796076.392758 quoted in the original description and in my first review. The 2^64 difference is exact.

The `count > bc_height` bound is confirmed a real defect and filed separately as #140. The `coinbase_amount - tx_fee_amount` note stands but is unreachable on a validated chain; the reasoning is in the resolved thread.

PR #128, opened 2026-08-15
fix(wallet): sweep rebuilt the transaction with a fee its destinations were not sized for

### LINKED ISSUE #127: sweep_all fails sanity_check ("Total received … expected …") on wallets with many outputs
## Summary

`sweep_all` (and any sweep that consolidates many outputs) aborts before sending with an internal error:

```
Error: internal error: Total received by <addr>: 13645.265497140612, expected 13645.268264670612
```

The wallet builds the sweep, then rejects its own transaction in `sanity_check`. The "received" amount is short of "expected" by a small amount (here `0.002767530000` XNV). No transaction is broadcast; the wallet is simply unable to sweep.

This is **not** ring-size specific and **not** introduced by the HF14 work. It reproduces on both ring-16 (HF14 testnet) and ring-5 transactions, and the responsible code predates the CLSAG/BP+ changes. It is a latent bug that surfaces whenever a wallet accumulates enough outputs (for example a mining wallet, or any wallet doing a large consolidation).

## Reproduction

1. Use a wallet with a large number of unspent outputs (reproduced reliably at ~200 coinbase outputs; a single 197-input sweep is enough).
2. Run `sweep_all <own_address>`, or the `sweep_all` RPC with `do_not_relay: true` (it fails at build time, so nothing needs to be broadcast).
3. The wallet throws `internal error: Total received by ... expected ...` and the sweep never completes.

Observed on wallet build `v0.3.0.0` (HF14 testnet, both wallet-rpc and CLI). The offending code is also present on mainnet builds (0.2.x), so mainnet wallets with many outputs are affected too.

## Root cause

In `wallet2::create_transactions_all` (`src/wallet/wallet2.cpp`), the sweep sizes its single destination inside the fee-adjustment `do { ... } while (needed_fee > test_ptx.fee)` loop. At convergence, `test_ptx` is a correct sweep: the destination is sized to `Sum(inputs) - F`, the embedded fee is `F` (= `test_ptx.fee`), and change is 0.

The loop exits as soon as the *recomputed* fee `needed_fee` (call it `F'`) drops to `<= test_ptx.fee` (`F`). Because the initial fee is derived from `estimate_tx_weight` and the recomputed fee is derived from the actual built transaction, `F'` is typically **strictly less** than `F`. The two adjacent lines after the loop then disagree about which fee to use:

```cpp
} while (needed_fee > test_ptx.fee);
...
tx.tx = test_tx;
tx.ptx = test_ptx;
tx.weight = get_transaction_weight(test_tx, txBlob.size());
tx.outs = outs;
tx.needed_fee = needed_fee;          // stores F' (the lower recomputed fee)
accumulated_fee += test_ptx.fee;     // uses F (the fee the tx was actually built with)
```

`tx.needed_fee` is set to `F'`, but the destination was sized for `F`. The subsequent real-signing reconstruction pass (the loop right after `hwdev.set_mode(hw::device::TRANSACTION_CREATE_REAL)`, which correctly rebuilds each tx for hardware signing) rebuilds the transaction with `tx.needed_fee = F'` while keeping the `F`-sized destination. The `F - F'` difference reappears as a **spurious change output**, so the sweep destination receives `Sum(inputs) - F` while `sanity_check` expects `Sum(inputs) - F'`. The shortfall equals exactly the spurious change (`F - F'`).

Debug-log evidence for one failing sweep (197 inputs, single tx):

```
Done creating 1 transactions, 0.166051800000 total fee, 0.000000000000 total change   (F, change 0)
Transaction 1/1 ...: sending 13645.428781410612 in 197 outputs to 1 destination(s),
                     including 0.163284270000 fee, 0.002767530000 change               (F', plus spurious change)
```

`0.166051800000 - 0.163284270000 = 0.002767530000`, which matches the `sanity_check` shortfall exactly.

## Suggested fix

Make the real-signing reconstruction use the same fee the destination was sized for. Minimal change:

```cpp
tx.needed_fee = test_ptx.fee;   // was: tx.needed_fee = needed_fee;
```

This rebuilds the final tx with fee `F`, no spurious change, and `sanity_check` passes. The tradeoff is that the sweep pays `F` (the estimate-based fee) rather than the marginally lower `F'`, an overpayment of `F - F'` (here ~0.0028 XNV on a ~0.163 XNV fee).

A cleaner alternative that also charges the tighter fee `F'`: in the reconstruction pass, re-size the destination(s) to `Sum(inputs) - F'` (add the freed `F - F'` back to the sweep destination) so change stays 0 at the lower fee.

## Notes

- Affects `sweep_all`, `sweep_below`, and any transfer the wallet routes through `create_transactions_all` with enough inputs to make `estimate_tx_weight` diverge from the final tx weight.
- Ring-size independent (confirmed on ring 5 and ring 16); worse with more inputs, since the estimate error scales with input count.
- Pre-existing code (2018-2020 era), unrelated to the HF14 CLSAG/BP+ changes.

mmokhi:
Closes #127.

The fee loop settles on a transaction built with fee F, then stores the fee recomputed from that transaction, F'. The loop exits when F' is at or below F, and it is usually below. The signing pass then rebuilds from the stored fee while the destinations are still sized for F, so F - F' comes back as change nobody asked for. A sweep has no change to absorb it, so sanity_check refuses the wallet's own transaction and the sweep never gets built.

Now it stores the fee the transaction was built with. That is what upstream does at both sites, and accumulated_fee on the next line was already using it, so the two stop disagreeing. create_transactions_2 had the same divergence and is fixed with it, it just has change to hide the difference in. It came in with dbec0a5, the Masari rebase, so it looks like a porting artifact rather than a decision.

Reproduced on a private testnet mined past the HF14 fork, 1101 blocks to one wallet, same height and same unlocked balance on both runs:

    before   ERROR  Total received 255100.829046170979, expected 255100.848475510979
    after    OK     2 transactions

The trigger is the per KB rounding in calculate_fee, kB = (bytes + 1023) / 1024. F and F' differ only when the estimate and the built transaction land in different KB buckets, so it turns on transaction size, not on the transaction format or the fork.

It fails before the fork too, which I checked rather than assumed. Sweeping one wallet at ten different below_amount thresholds, so each sweep pulls a different number of inputs, on chains below and above the fork:

                        before      after
    pre fork, HF13      6 failures  0 failures
    post fork, HF14     fails       passes

Pre fork is HF13, the format mainnet runs today. Two earlier pre fork runs of mine passed and that was one bucket of luck, not the format protecting anything.

So this is not gated on HF14. A mainnet wallet with enough outputs can hit it today.

On effects, since it is a money path. tx.needed_fee has two consumers, both the fee argument to transfer_selected_rct in the rebuild. The fee can only move up, never down, since the loop exits when F' is at or below F, so it cannot underpay and get rejected. The transaction carries one output fewer with the spurious change gone. The cost is paying F instead of F', 0.0028 on a 0.166 fee in the report.

---

Maintainer note: the author is unavailable, so this branch was rebased onto master and the two review points settled in place. The comment blocks at both sites were trimmed to one line each. The second commit, the `relay-sanity-checks` toggle, was split out: its root cause is still open and is tracked in #139, so it does not ride along on a squash that closes #127. The dropped commit is attached as a patch on #139.

### DIFF
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -9476,7 +9476,8 @@ std::vector<wallet2::pending_tx> wallet2::create_transactions_2(std::vector<cryp
         tx.ptx = test_ptx;
         tx.weight = get_transaction_weight(test_tx, txBlob.size());
         tx.outs = outs;
-        tx.needed_fee = needed_fee;
+        // the fee the tx was built with; the rebuild below sizes change against it
+        tx.needed_fee = test_ptx.fee;
         accumulated_fee += test_ptx.fee;
         accumulated_change += test_ptx.change_dts.amount;
         adding_fee = false;
@@ -9836,7 +9837,8 @@ std::vector<wallet2::pending_tx> wallet2::create_transactions_from(const crypton
       tx.ptx = test_ptx;
       tx.weight = get_transaction_weight(test_tx, txBlob.size());
       tx.outs = outs;
-      tx.needed_fee = needed_fee;
+      // the fee the tx was built with; a sweep has no change to absorb a lower one
+      tx.needed_fee = test_ptx.fee;
       accumulated_fee += test_ptx.fee;
       accumulated_change += test_ptx.change_dts.amount;
       if (!unused_transfers_indices.empty() || !unused_dust_indices.empty())


src/simplewallet/simplewallet.cpp
```diff
@@ -2935,6 +2946,7 @@ simple_wallet::simple_wallet()
                                   "print-ring-members <1|0>\n "
                                   "  Whether to print detailed information about ring members during confirmation.\n "
                                   "store-tx-info <1|0>\n "
+                                  "relay-sanity-checks <1|0>\n "
```
sn1f3rt: The new entry lands between `store-tx-info <1|0>` and its description line, so `help set` now renders:

```
store-tx-info <1|0>
relay-sanity-checks <1|0>
  Whether to store outgoing tx info (destination address, payment ID, tx secret key) for future reference.
```

`store-tx-info` loses its description and this option inherits the wrong one. Move the entry below line 2950 and give it its own line, e.g. `Whether to ask the daemon to run its relay sanity checks on submitted txes. Session only, not saved.`

src/wallet/wallet2.cpp
```diff
@@ -9476,7 +9477,12 @@ std::vector<wallet2::pending_tx> wallet2::create_transactions_2(std::vector<cryp
         tx.ptx = test_ptx;
         tx.weight = get_transaction_weight(test_tx, txBlob.size());
         tx.outs = outs;
-        tx.needed_fee = needed_fee;
+        // The fee the transaction was actually built with, not the lower one
```
sn1f3rt: Checked this against upstream: monero master uses `tx.needed_fee = test_ptx.fee;` at both sites (wallet2.cpp:10850 and :11288), so the parity claim holds. Coverage is complete too, `needed_fee` only exists in these two functions and every sweep entry point funnels through `create_transactions_from`.

On the comment itself: five lines for a one-token change, restated again at 9845. House style is minimal comments carrying only the essential why. Both sites would read fine as `// the fee the tx was built with; the rebuild below sizes change against it`.

src/wallet/wallet2.cpp
```diff
@@ -6286,7 +6287,7 @@ void wallet2::commit_tx(pending_tx& ptx)
     COMMAND_RPC_SEND_RAW_TX::request req;
     req.tx_as_hex = epee::string_tools::buff_to_hex_nodelimer(tx_to_blob(ptx.tx));
     req.do_not_relay = false;
-    req.do_sanity_checks = true;
+    req.do_sanity_checks = m_relay_sanity_checks;
```
sn1f3rt: This is the line that spends the privacy, so worth pinning the root cause before it ships.

Your own testing says the median rule could not be tripped naturally even at 11916 rct outputs, and the passing test forced `tx_sanity_check` to fail. So what is proven is that the override works, not that the median rule is what blocks those two wallets.

The reported numbers are the part I would chase: ring offsets topping out at 17254 of a 31474 pool means no decoy came from the recent 45% of the chain. Gamma selection clusters decoys near the tip, so that shape looks like `get_outs` fell off the gamma path, e.g. `get_rct_distribution` returning false at wallet2.cpp:7569 and dropping to histogram-based triangular selection, which is flat and would produce exactly this. If that is what happened, the rings really are bad and the daemon is correctly refusing them, and this switch turns "your rings are bad" into "your rings are bad and now they are on chain permanently".

Worth noting in the change's favour: the daemon does not gate `do_sanity_checks` on restricted RPC (core_rpc_server.cpp:1207), so any peer can already bypass it with raw hex. The check protects the sender's own ring, not the network, so the privacy-trade framing is right. Also note upstream hardcodes this to `true` (wallet2.cpp:7606) and has never exposed it, so this is a deliberate Nerva divergence rather than a port.

src/simplewallet/simplewallet.cpp
```diff
@@ -2280,6 +2280,17 @@ bool simple_wallet::set_store_tx_info(const std::vector<std::string> &args/* = s
   return true;
 }
 
+bool simple_wallet::set_relay_sanity_checks(const std::vector<std::string> &args/* = std::vector<std::string>()*/)
```
sn1f3rt: Once this is off, nothing warns again. The transfer confirmation prompt is the moment the privacy is actually spent, and a user who set this ten minutes ago will not connect the two. Consider a line in the confirmation when the setting is off.

Verified the rest of the lifetime claim: both the CLI (`m_wallet = std::move(rc.first)`) and wallet-rpc `on_open_wallet` build a fresh `wallet2`, so it genuinely does not survive a reopen.

src/wallet/wallet_rpc_server.cpp
```diff
@@ -3082,6 +3082,22 @@ namespace tools
       return false;
     }
   }
+  bool wallet_rpc_server::on_set_relay_sanity_checks(const wallet_rpc::COMMAND_RPC_SET_RELAY_SANITY_CHECKS::request& req, wallet_rpc::COMMAND_RPC_SET_RELAY_SANITY_CHECKS::response& res, epee::json_rpc::error& er, const connection_context *ctx)
+  {
+    if (!m_wallet) return not_open(er);
```
sn1f3rt: `!m_wallet` is checked before `m_restricted` here, the reverse of the neighbouring handlers (`on_auto_refresh`, `on_set_daemon`). A restricted caller learns whether a wallet is open before being denied. Trivial, but #122 just went through this file for restricted-mode hardening, so the consistency is cheap.

sn1f3rt (commented):
Commit 1 is correct and I would merge it as is, modulo the comment trim. Verified the diagnosis against the code, confirmed upstream parity at both sites, confirmed the fee can only move up so there is no underpay path, and syntax-compiled all three changed translation units clean.

Commit 2 is a separate concern: `Closes #127` covers only the first commit, so merging as is means the #127 changelog entry silently ships a privacy toggle. Suggest splitting it into its own issue and PR. That also lets the sweep fix land now while the decoy-selection question in the wallet2.cpp:6290 thread gets settled.

sn1f3rt:
The author is unavailable, so I took the branch the rest of the way rather than let the sweep fix sit. Rebased onto master, trimmed the comment at both sites to one line, and dropped the second commit. What ships here is the one-token fee change and nothing else.

Re-verified the fee fix from scratch on the rebased branch rather than trusting the earlier pass:

- `transfer_selected_rct` assigns `ptx.fee = fee` verbatim (wallet2.cpp:8405), so on loop exit `test_ptx.fee` is the fee the transaction was built with and `needed_fee` is the recomputed one, which the loop condition keeps at or below it.
- In `create_transactions_from` the destinations are re-sized inside the loop from `available_for_fee - needed_fee` at the top of each iteration, so on the last iteration they are sized against exactly the fee the build used. In `create_transactions_2` the destinations are fixed and change absorbs the difference. Either way the rebuild now reproduces the transaction the loop converged on.
- `sanity_check` in the sweep is handed `inputs - ptx.fee`, so with the old code it expected the lower fee while the destination had been sized for the higher one. The shortfall is F - F' exactly, which is what the report shows.
- `TX::needed_fee` is written and read only in these two functions, so coverage is complete.
- The "take fee from paid amount" branch at wallet2.cpp:9448 assigns `test_ptx.fee = needed_fee` before either store, so that path is bit-identical before and after.
- The fee can only move up, and the transaction is one output smaller with the spurious change gone, so there is no underpay or weight risk.

Two corrections to my own earlier review, both on the second commit.

The `wallet2.cpp:6290` thread blamed `get_rct_distribution` falling back to triangular selection, which I described as flat. That is wrong: triangular there is `sqrt`-biased toward recent outputs and passes the median rule comfortably. I simulated all four selection paths in `get_outs` against the reported chain shape using this tree's constants, and none of them reproduces offsets with a median at 30% and a maximum at 55% of the pool. Gamma puts the median around 80%, and it still does on a very young chain, since the rejection loop in `gamma_picker::pick` discards ages longer than the chain and the survivors stay near the tip. The reported shape only reproduces if the wallet drew from a pool roughly 40% smaller than the one the daemon measured.

So the conclusion stands but for a better reason, and it is now filed as #139 with the numbers and the log lines that would settle it. The dropped commit is attached as a patch there.

The `wallet_rpc_server.cpp:3087` thread is withdrawn outright. I claimed the `!m_wallet` before `m_restricted` order was reversed relative to its neighbours; `on_set_daemon`, the handler immediately above it in the map, uses the identical order. There was nothing to fix.

The `simplewallet.cpp:2949` help-text bug was real and is still worth carrying over to whatever revives the toggle: the new entry landed between `store-tx-info <1|0>` and its description, so `store-tx-info` lost its description and the new option inherited it.

Thanks @mmokhi, the diagnosis and the write-up on the fee fix were exact, down to which line the two consumers disagree on.

src/wallet/wallet2.cpp
sn1f3rt: Trimmed to one line at both sites. Also re-confirmed the parity claim directly this time rather than from memory: monero master has `tx.needed_fee = test_ptx.fee;` at wallet2.cpp:10856 and :11294, in the same two functions with the same loop shape.

src/wallet/wallet2.cpp
sn1f3rt: Withdrawing the mechanism I proposed here, it does not hold. Simulating this tree's own picker shows gamma puts the ring median around 80% of the pool and passes at every chain age, and the histogram fallback is sqrt-biased toward recent outputs so it passes too. The conclusion stands for a better reason and moves to #139 with the numbers. The dropped commit is attached as a patch there.

src/wallet/wallet_rpc_server.cpp
sn1f3rt: Withdrawn, this was my error. `on_set_daemon`, the handler immediately above it in the map, checks `!m_wallet` before `m_restricted` exactly the same way. There was nothing inconsistent.

src/simplewallet/simplewallet.cpp
sn1f3rt: Still a real bug, but the code left this PR with the second commit. Carried to #139 so it is not lost.

src/simplewallet/simplewallet.cpp
sn1f3rt: Not addressed, and the code left this PR with the second commit. Carried to #139 along with the related point that the rejection message does not point a stuck user at the setting either.

PR #129, opened 2026-08-15
feat: enhance network privacy, RPC API, mining, and prepare structures

XelisVault:
Summary
This PR adds 15 features to the Nerva codebase, organized as 9 atomic commits. Each commit is self-contained and can be reviewed independently.
Changes
Network Privacy (ready to use)
* Dandelion++ stem propagation (--dandelion++): Activates the dormant Dandelion++ infrastructure to forward transactions via a single stem peer, breaking the link between originator and transaction
* Tor v3 onion auto-config (--anonymous-inbound auto): Automatically creates a v3 onion service via the Tor control port, eliminating manual torrc editing
RPC and API Enhancements (ready to use)
* Enhanced fee estimator: get_fee_estimate now returns p10/p50/p90 percentiles and 100-block fee history
* OpenAPI 3.0 spec generator: Python script that generates docs/openapi.yaml for SDK auto-generation
* HTTP webhooks (--block-webhook, --reorg-webhook): JSON POST notifications as alternative to subprocess-based --block-notify
New Modules (compiled, need daemon wiring)
* WebSocket subscription server: Push real-time new_block/new_tx/reorgnotifications to connected clients
* Miner IPC API: Unix domain socket for NervaOne GUI to control the miner without HTTP RPC
HF15 Feature Structures (dormant until hard fork)
* View Tags: 1-byte per-output tag for 8x faster wallet scanning
* Burn-to-prioritize: EIP-1559-style fee burning for deflationary priority
* Payment ID v2: Extended 32-byte encrypted payment IDs
* UTXO snapshot export/import: Bootstrap in minutes instead of hours
* P2P zstd compression: ~40% bandwidth reduction on block/tx relay
Design Documents
* Multisig HF14 fix: Detailed plan to port CLSAG multisig from Monero (PRs #8123, #8234, #8580)
* Ledger HF14 workaround: Options for unblocking Ledger users on HF14
See docs/FEATURE_ROADMAP.md for the full implementation status, design decisions, and compilation notes for each feature.
Review Process
Each commit is atomic and self-contained:
1. feat(network): activate Dandelion++ stem propagation
2. feat(notify): add HTTP webhook notifications
3. feat(rpc): enhance get_fee_estimate with percentiles
4. feat(network): Tor v3 onion service auto-configuration
5. feat(docs): add OpenAPI 3.0 spec generator
6. feat(rpc): add WebSocket subscription server
7. feat(miner): add IPC server for NervaOne GUI
8. feat: add HF15 feature structures, UTXO snapshot, P2P zstd
9. docs: update README with enhanced features
Recommended review order: review each commit independently, attempt a local build after each, run contrib/hf14checks/tests, and test on testnet before mainnet.
Backward Compatibility
* All new features are opt-in (disabled by default)
* The enhanced get_fee_estimateresponse is backward compatible (new fields are additional)
* The existing --block-notify and --reorg-notify subprocess mechanisms are unchanged
* No consensus rules are changed (HF15 features are dormant structures only)

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 ## Table of Contents
 
 - [About](#about)
+- [Enhanced Features](#enhanced-features)
 - [Installing](#installing)
 - [Contributing](#contributing)
 - [Help](#help)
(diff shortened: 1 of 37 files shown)


mmokhi:
Not merging this, and I'm closing it rather than asking for changes.

15 unrelated features in 37 files can't be reviewed or reverted as a unit, no CI ran on it, and by your own description a good part is code that isn't wired to anything.

To be clear on the rest of it: there is no HF15. Nothing is planned, nothing is scoped, no date, no feature list. HF14 isn't even on mainnet yet. So "HF15 feature structures" are structures for something that does not exist, and a PR is not where a fork's contents get decided. HF14 had a tracking issue, a testnet and measured work behind every change in it.

The Dandelion++ part is broken as written. CRYPTONOTE_DANDELIONPP_EMBARGO_TIMEOUT is defined and never used, so there's no embargo and no fluff. Local transactions all share the nil source uuid, so every transaction I originate goes to the same peer for the whole epoch and that peer learns exactly which ones are mine. If it drops them, nothing rebroadcasts. That is worse than flooding, and the comment right above it says it defeats passive observers.

view_tags.h doesn't match Monero's construction either. Monero hashes the derivation directly, this hashes it into a seed and hashes again. That's a consensus-visible byte, it has to be a port, not a re-invention.

Burn-to-prioritize changes monetary policy and makes the mempool pay-to-win. 32 byte payment ids fingerprint the people who use them, which is why Monero dropped them. Neither belongs in a privacy coin, and neither gets decided this way.

The zstd code is fine, bounds are correct.

If you are generating this with a tool, please audit the output before submitting. An embargo constant that nothing reads, and a view tag that doesn't match the construction it claims to implement, are exactly what unaudited generation looks like. Both read fine and neither does what its comment says.

If you want to contribute: one feature per PR, nothing dormant, nothing unwired, and nothing that assumes a hard fork nobody has planned.

PR #130, opened 2026-08-16
docs: add OpenAPI 3.0 spec generator for the daemon RPC API

XelisVault:
The Nerva daemon exposes ~43 REST endpoints and ~41 JSON-RPC methods, but there is no machine-readable API specification. This makes it difficult to generate SDK clients or integrate with API testing tools (Postman, Insomnia, swagger-codegen).

This commit adds a Python script that parses src/rpc/core_rpc_server.h and generates docs/openapi.yaml, an OpenAPI 3.0 specification covering all daemon RPC endpoints.

The script uses regex to extract the endpoint list from the MAP_URI_AUTO_JON2 / MAP_URI_AUTO_BIN2 / MAP_JON_RPC macros in the header, then builds the OpenAPI spec with:
- All 43 REST endpoints (including 17 restricted and 9 binary)
- The /json_rpc endpoint listing all 41 JSON-RPC methods
- 19 reusable schema components for the most common response types (GetInfoResponse, BlockHeader, TransactionEntry, etc.)
- Proper tagging (Network, Blocks, Transactions, Mempool, Mining, Peers, Nerva-specific, Restricted, etc.)
- Server URLs for mainnet (17566), testnet (18566), stagenet (19566)

Usage:
    python3 scripts/openapi/generate_openapi.py

Requires PyYAML (pip install pyyaml).

No C++ files are modified. The generated spec is documentation-only and does not affect the daemon build.


scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,705 @@
...
+        method = m.group(1)
+        handler = m.group(2)
+        command = m.group(3)
+        condition = m.group(4) or ""
+        restricted = "m_restricted" in condition
+        json_rpc_methods.append({
+            "method": method,
+            "handler": handler,
+            "command": command,
+            "restricted": restricted,
```
mmokhi: This is collected and then never emitted. 12 json_rpc methods are gated on !m_restricted in core_rpc_server.h: add_peer, banned, flush_cache, flush_txpool, get_alternate_chains, get_bans, get_coinbase_tx_sum, get_connections, prune_blockchain, relay_tx, set_bans, sync_info. The REST side marks all 17 of its restricted endpoints correctly, so the spec is inconsistent with itself.

docs/openapi.yaml
```diff
@@ -0,0 +1,1253 @@
...
+                description: Response schema (see COMMAND_RPC_* in core_rpc_server_commands_defs.h)
+  /json_rpc:
+    post:
+      tags:
+      - JSON-RPC
+      summary: JSON-RPC 2.0 endpoint
+      description: 'Submit a JSON-RPC 2.0 request. The `method` field selects the daemon RPC command.
+
+
+        Available methods (41 total):
```
mmokhi: All 41 listed flat with no distinction. set_bans, prune_blockchain, flush_txpool and add_peer sitting next to get_info tells an integrator they are safe to expose on a public node. They are not.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,705 @@
...
+        sys.exit(1)
+
+    rest_endpoints, json_rpc_methods = parse_endpoints_from_header()
+
+    spec, method_names = build_openapi_spec(rest_endpoints, json_rpc_methods)
+
+    # Ensure docs directory exists
+    OUTPUT_FILE.parent.mkdir(parents=True, exist_ok=True)
+
+    with open(OUTPUT_FILE, "w") as f:
```
mmokhi: The generator and its output are both committed with nothing keeping them in sync. First person to add an rpc method makes docs/openapi.yaml wrong and nothing catches it.\n\nMy preference is to drop the yaml and let people run this when they need it, I would rather not carry generated files in the tree. Otherwise it needs a CI job that regenerates and fails on drift.\n\nAlso PyYAML is not in any requirements file.

docs/openapi.yaml
mmokhi: /json_rpc has no operationId. Every one of the 41 methods goes through this endpoint and codegen names the client method from it.

mmokhi (changes requested):
Three things before this goes in, plus a nit. Details inline.

I ran the script against master and diffed its output against the committed yaml, and checked the endpoint counts, ports and .bin handling against core_rpc_server.h. The points below are what came out of that.

XelisVault:
Yes sorry 😅 it’s fixed now ;)

mmokhi (approved):
Thanks 👍

All four addressed. I re-ran the script against master to check: restricted list matches the header exactly, operationId is there, the yaml is gone and gitignored, requirements file added.

LGTM ✅

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+    public_methods = sorted(m["method"] for m in json_rpc_methods if not m["restricted"])
+    restricted_methods = sorted(m["method"] for m in json_rpc_methods if m["restricted"])
+
+    desc = (
+        "Submit a JSON-RPC 2.0 request. The `method` field selects the daemon RPC command.\n\n"
+        f"Public methods ({len(public_methods)} total) — safe to expose on a public node:\n"
+        + ", ".join(f"`{m}`" for m in public_methods)
+        + "\n\n"
+        f"Restricted methods ({len(restricted_methods)} total) — require the daemon to be started "
+        "without --restricted-rpc, or with --rpc-login authentication. Do NOT expose these on a public node:\n"
```
sn1f3rt: This is factually wrong, and it is the most serious problem in the PR.

`--rpc-login` does **not** lift restricted mode. `m_restricted` is assigned exactly once, at `src/rpc/core_rpc_server.cpp:227`, from the `restricted` argument passed to `init()`. That argument comes from `arg_restricted_rpc` (`src/daemon/daemon.cpp:80`) or is hardcoded `true` for the `--rpc-restricted-bind-port` server (`src/daemon/daemon.cpp:88`). `rpc_config->login` is independent HTTP digest authentication and never touches it.

So on a daemon started with `--restricted-rpc`, adding `--rpc-login` still leaves `set_bans`, `flush_txpool`, `relay_tx`, `prune_blockchain`, `add_peer` etc. unavailable. As written this sentence reads both ways and both are dangerous: it tells an operator that an authenticated restricted node exposes admin methods, and it implies `--rpc-login` is a substitute for `--restricted-rpc`.

Suggest: "require the daemon to be started without `--restricted-rpc` and without `--rpc-restricted-bind-port`. `--rpc-login` provides authentication but does not re-enable restricted methods."

Same text also appears at line 257.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+        "openapi": "3.0.3",
+        "info": {
+            "title": "Nerva Daemon RPC API",
+            "description": (
+                "REST and JSON-RPC API for the Nerva (XNV) daemon.\n\n"
+                "The daemon exposes two types of endpoints:\n"
+                "1. REST-style endpoints (e.g. /get_info, /get_transactions)\n"
+                "2. JSON-RPC 2.0 methods under /json_rpc\n\n"
+                "Endpoints marked as restricted require the daemon to be started "
+                "without --restricted-rpc, or with --rpc-login authentication.\n\n"
```
sn1f3rt: Same incorrect `--rpc-login` claim as line 238 — this copy lands in `info.description`, so it is the first thing anyone importing the spec reads. Please fix both.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+                "Default ports: mainnet 17566, testnet 18566, stagenet 19566."
+            ),
+            "version": "1.0.0",
+            "contact": {
+                "name": "Nerva Project",
+                "url": "https://nerva.one",
+            },
+            "license": {
+                "name": "MIT",
```
sn1f3rt: Wrong license. Nerva is BSD-3-Clause, not MIT — see `LICENSE` lines 1-20 (three-clause redistribution terms, Nerva/Monero/Masari copyright holders).

This matters beyond cosmetics: `info.license` is propagated into the package metadata of clients produced by swagger-codegen / openapi-generator, so this would mislabel every downstream SDK.

```
"license": {
    "name": "BSD-3-Clause",
    "url": "https://opensource.org/licenses/BSD-3-Clause",
},
```

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+        },
+    }
+
+
+def main():
+    if not SERVER_HEADER.exists():
+        print(f"Error: {SERVER_HEADER} not found", file=sys.stderr)
+        sys.exit(1)
+
+    rest_endpoints, json_rpc_methods = parse_endpoints_from_header()
```
sn1f3rt: The parse result is never sanity-checked, so the generator fails silently.

I copied `core_rpc_server.h`, renamed `MAP_URI_AUTO_JON2` -> `MAP_URI_AUTO_JSON2` and `MAP_JON_RPC` -> `MAP_JSON_RPC` (a plausible refactor), and re-ran the script. It printed:

```
Generated: .../docs/openapi.yaml
  REST endpoints: 9 (0 restricted)
  JSON-RPC methods: 0 (0 public, 0 restricted)
```

and exited **0**. The `BIN2` macros still matched, so it produced a plausible-looking but almost empty spec and reported success.

Because the output is gitignored and no workflow in `.github/workflows/` runs Python, nothing would ever catch this. Please assert the parse found something and exit non-zero otherwise, e.g.:

```python
if not rest_endpoints or not json_rpc_methods:
    print("Error: no endpoints parsed from %s - macro format may have changed" % SERVER_HEADER, file=sys.stderr)
    sys.exit(1)
```

Minimum-count assertions would be even better.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+            if ep["restricted"]:
+                description += " Restricted: requires non-restricted RPC."
+            if ep["command"] in NERVA_SPECIFIC or "donate" in path:
+                description += " Nerva-specific (not in Monero)."
+
+        tags = [tag]
+        if ep["restricted"]:
+            tags.append("Restricted")
+
+        operation = {
```
sn1f3rt, line 1133: No operation defines a `requestBody`, and none defines `parameters` — 43 of the 44 generated operations document responses only. I confirmed this on the generated spec:

```
ops without requestBody: 43
total ops: 44
```

But `/get_transactions` (`txs_hashes`, `decode_as_json`), `/send_raw_transaction` (`tx_as_hex`, `do_not_relay`, `do_sanity_checks`), `/is_key_image_spent` (`key_images`), `/get_outs` (`outputs`) and `/pop_blocks` (`nblocks`) all take request bodies.

The PR's stated goal is generating SDK clients and Postman collections. A client generated from this spec would POST an empty body to every one of those endpoints, which makes the generated SDK non-functional for anything that takes arguments. This is the biggest functional gap here — even hand-written request schemas for the handful of endpoints that already have hand-written response schemas would be a large improvement.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+    if "peer" in path or "connection" in handler.lower() or "ban" in handler.lower():
+        return "Peers"
+    if "log" in path:
+        return "Logging"
+    if "daemon" in handler.lower() or "save" in path or "update" in path or "limit" in path:
+        return "Daemon"
+    return "Network"
+
+
+def categorize_json_rpc(method):
```
sn1f3rt: `categorize_json_rpc()` is dead code — it is defined here and never called anywhere in the file (only `categorize_endpoint()` is used, at line 304).

The consequence is visible in the output: JSON-RPC methods are emitted only as a comma-separated list of names inside the `/json_rpc` description and carry no tags at all. That in turn means four of the tags declared at lines 276-290 — `Fees`, `Outputs`, `Hard Forks`, `Version` — are never applied to any operation in the generated spec.

Either wire this up (e.g. by emitting per-method operations, see the schema comment on line 200) or drop the function and the four unused tags.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+
+
+def build_schema_for_command(command):
+    """
+    Build an OpenAPI schema reference for a COMMAND_RPC_* struct.
+    For now, we return a generic object since auto-parsing the C++ structs
+    from commands_defs.h would require a full C++ parser.
+    The most common endpoints have hand-written schemas in the components section.
+    """
+    known_schemas = {
```
sn1f3rt: 8 of these 15 entries can never match, because `build_schema_for_command()` is only ever called for REST endpoints (line 330) and these commands are registered exclusively under `/json_rpc`:

`COMMAND_RPC_GET_PER_KB_FEE_ESTIMATE`, `GET_BLOCK_HEADER_BY_HEIGHT`, `GET_BLOCK_HEADERS_RANGE`, `GET_LAST_BLOCK_HEADER`, `GET_BLOCK`, `HARD_FORK_INFO`, `GET_VERSION`, `GET_GENERATED_COINS`.

Resolving refs transitively over the generated spec, 9 of the 19 component schemas end up unreachable from any operation:

```
orphaned: ['BlockHeader', 'BlockHeaderByHeightResponse', 'BlockHeadersRangeResponse',
           'FeeEstimateResponse', 'GetBlockResponse', 'GetGeneratedCoinsResponse',
           'GetVersionResponse', 'HardForkInfoResponse', 'LastBlockHeaderResponse']
```

Codegen tools commonly prune unreferenced components, so roughly half the hand-written schema work in this file is discarded at generation time. The fix is to give the JSON-RPC methods real operations rather than a name list in a description.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+            {"name": "Outputs", "description": "Output queries"},
+            {"name": "Hard Forks", "description": "Hard fork information"},
+            {"name": "Version", "description": "Daemon version info"},
+            {"name": "Logging", "description": "Log level management"},
+            {"name": "Daemon", "description": "Daemon control and maintenance"},
+            {"name": "Nerva-specific", "description": "Endpoints specific to Nerva (not in Monero)"},
+            {"name": "Restricted", "description": "Endpoints requiring non-restricted RPC"},
+        ],
+        "paths": {},
+        "components": {
```
sn1f3rt: No `securitySchemes` is emitted, even though the spec text at lines 238 and 257 tells users about `--rpc-login`. Generated clients therefore have no way to authenticate against a node that requires it.

The daemon uses HTTP digest auth, so:

```python
"components": {
    "securitySchemes": {
        "rpcLogin": {"type": "http", "scheme": "digest",
                     "description": "Enabled by --rpc-login user:pass"},
    },
    "schemas": build_component_schemas(),
},
```

plus an optional top-level `security` entry.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+        if m["method"] not in seen:
+            seen.add(m["method"])
+            unique_methods.append(m)
+
+    return rest_endpoints, unique_methods
+
+
+# Human-readable descriptions for the most common endpoints.
+# Endpoints not in this dict get a generic description.
+ENDPOINT_DESCRIPTIONS = {
```
sn1f3rt: Keying this by path rather than by handler means every alias endpoint silently loses its description. 7 paths are affected:

`/getheight`, `/getinfo`, `/gettransactions`, `/sendrawtransaction`, `/getblocks.bin`, `/getblocks_by_height.bin`, `/gethashes.bin`

They all fall back to `"RPC endpoint handled by on_… (COMMAND_RPC_…)."`. Overall 20 of 44 generated paths use the fallback description, and 7 of those have a perfectly good hand-written description sitting right here under the non-alias spelling.

Keying on `ep["handler"]` instead of `path` fixes all 7 with no extra content to write.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+                "votes": {"type": "integer"},
+                "threshold": {"type": "integer"},
+                "voting": {"type": "integer", "description": "Version being voted for"},
+                "state": {"type": "integer"},
+                "earliest_height": {"type": "integer", "format": "uint64", "description": "Height at which the next fork activates"},
+                "status": {"type": "string", "example": "OK"},
+                "untrusted": {"type": "boolean"},
+            },
+        },
+        "GetVersionResponse": {
```
sn1f3rt: Doesn't match the struct. `COMMAND_RPC_GET_VERSION::response_t` in `src/rpc/core_rpc_server_commands_defs.h` serializes:

```cpp
uint32_t version;
bool release;
```

plus `status`/`untrusted` from `rpc_response_base`. `release` is missing here.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+            "type": "object",
+            "properties": {
+                "version": {"type": "integer", "description": "Daemon version as an integer"},
+                "status": {"type": "string", "example": "OK"},
+                "untrusted": {"type": "boolean"},
+            },
+        },
+        "GetGeneratedCoinsResponse": {
```
sn1f3rt: `untrusted` is not part of this response. Unlike almost every other command, `COMMAND_RPC_GET_GENERATED_COINS::response_t` does **not** inherit `rpc_response_base`:

```cpp
struct response_t
{
  std::string status;
  uint64_t coins;

  BEGIN_KV_SERIALIZE_MAP()
    KV_SERIALIZE(status)
    KV_SERIALIZE(coins)
  END_KV_SERIALIZE_MAP()
};
```

Only `status` and `coins` are serialized, so the `untrusted` property should be dropped here.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+                "rpc_connections_count": {"type": "integer", "format": "uint64"},
+                "white_peerlist_size": {"type": "integer", "format": "uint64"},
+                "grey_peerlist_size": {"type": "integer", "format": "uint64"},
+                "mainnet": {"type": "boolean"},
+                "testnet": {"type": "boolean"},
+                "stagenet": {"type": "boolean"},
+                "nettype": {"type": "string", "example": "mainnet"},
+                "top_block_hash": {"type": "string"},
+                "cumulative_difficulty": {"type": "integer", "format": "uint64"},
+                "cumulative_difficulty_top64": {"type": "integer", "format": "uint64"},
```
sn1f3rt: `format: uint64` is not a defined OpenAPI format — the spec defines `int32` and `int64` for `type: integer`. Arbitrary format strings are legal but ignored by every generator, so all of these degrade to a plain integer.

That is mostly harmless, except for the fields that genuinely exceed the int64 / JS-safe-integer range — `cumulative_difficulty` and `cumulative_difficulty_top64` here, and `difficulty` on very long chains. Those need explicit handling (e.g. `type: integer, format: int64` where it fits, or a documented string representation where it doesn't) rather than a format token that tools discard.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+
+    rest_endpoints, json_rpc_methods = parse_endpoints_from_header()
+
+    spec = build_openapi_spec(rest_endpoints, json_rpc_methods)
+
+    # Ensure docs directory exists
+    OUTPUT_FILE.parent.mkdir(parents=True, exist_ok=True)
+
+    with open(OUTPUT_FILE, "w") as f:
+        yaml.dump(spec, f, default_flow_style=False, sort_keys=False, width=120)
```
sn1f3rt: `yaml.dump()` defaults to `allow_unicode=False`, so the em dashes in the `/json_rpc` description come out as literal escapes in the generated file:

```
\ total) \u2014 safe to expose on a public node:\n`decode_outputs`, ...
```

Valid YAML, but it renders badly in Swagger UI diffs and in the raw file. Pass `allow_unicode=True`, and add `encoding="utf-8"` to the `open()` on line 677 so the script doesn't depend on the platform default encoding.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+        }
+
+        if path not in spec["paths"]:
+            spec["paths"][path] = {}
+        spec["paths"][path][method] = operation
+
+    # Add the JSON-RPC endpoint
+    spec["paths"]["/json_rpc"] = {
+        "post": {
+            "tags": ["JSON-RPC"],
```
sn1f3rt, line 1163: The `JSON-RPC` tag isn't declared in the top-level `tags` list (lines 276-290). Legal, but tools that render tag descriptions will show it as an undocumented group. Add it alongside the others.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
...
+    "get_generated_coins", "get_min_version", "get_tx_pubkey",
+    "decode_outputs", "add_peer", "set_donate_level",
+}
+
+
+def categorize_endpoint(path, handler, command):
+    """Return the tag (category) for an endpoint."""
+    if command in NERVA_SPECIFIC or "donate" in path:
+        return "Nerva-specific"
+    if "transaction_pool" in path or "txpool" in path or "mempool" in handler.lower():
```
sn1f3rt: `"mempool" in handler.lower()` never matches — every pool handler in `core_rpc_server.h` is named `on_get_transaction_pool*` / `on_flush_txpool`. The `path` checks already cover these, so this clause can go.

Minor related note: the ordering here puts `"block" in path` ahead of the mining check, so a path like `/getblocktemplate` would be tagged `Blocks` rather than `Mining`. Currently moot since block-template is JSON-RPC only and JSON-RPC methods aren't tagged at all (see line 164), but it would bite if the JSON-RPC operations get built out.

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,693 @@
+#!/usr/bin/env python3
```
sn1f3rt, line 1: The file has a `#!/usr/bin/env python3` shebang but is committed with mode `100644`, so `./scripts/openapi/generate_openapi.py` won't run directly. Either `chmod +x` it or drop the shebang.

docs/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Generated by scripts/openapi/generate_openapi.py — do not commit
+openapi.yaml
```
sn1f3rt: Worth reconsidering: gitignoring the output means anyone who wants the spec needs a Python toolchain and PyYAML, which undercuts the "import into Postman / Insomnia / swagger-codegen" motivation in the PR description — those users are not necessarily building the daemon from source.

Committing `docs/openapi.yaml` and adding a small CI job that regenerates it and fails on a diff would deliver the stated benefit directly, and would also cover the silent-failure problem I flagged at `generate_openapi.py:670`.

sn1f3rt (changes requested):

XelisVault:
I have fixed everything ;)

scripts/openapi/generate_openapi.py
```diff
@@ -0,0 +1,1348 @@
...
+# Struct parsing: commands_defs.h and cryptonote_protocol_defs.h
+# ---------------------------------------------------------------------------
+
+STRUCT_HEAD_RE = re.compile(r"\bstruct\s+(\w+)\s*(?::\s*(?:public\s+)?(\w+)\s*)?\{")
+MEMBER_RE = re.compile(
+    r"^\s*([A-Za-z_]\w*(?:::[A-Za-z_]\w*)*(?:\s*<[^;=]+>)?)\s+([A-Za-z_]\w*)\s*;"
+    r"(?:\s*//\s*(.*?)\s*)?$"
+)
+# KV_SERIALIZE macros, longest names first so the alternation cannot truncate.
+KV_MACRO_RE = re.compile(
```
sn1f3rt, line 290: Non-blocking, and the only survivor from the last round — it is the same failure class as the routing-table gap, one level down.

The routing macros now have a total-count cross-check at line 264, but the KV macros don't. `KV_MACRO_RE` matches a fixed alternation of suffixes, so a macro whose suffix isn't in the list doesn't match **at all** — it is invisible rather than "unhandled", and the `raise ParseError(f"unhandled KV_SERIALIZE macro ...")` at line 394 never fires. The field is silently dropped from the schema.

Demonstrated by substituting one macro in `COMMAND_RPC_GET_HEIGHT`:

```
KV_SERIALIZE(hash)  ->  KV_SERIALIZE_HEX(hash)
```

```
exit=0
committed spec GetHeightResponse: ['hash', 'height', 'status', 'untrusted']
after KV_SERIALIZE_HEX  : ['height', 'status', 'untrusted']
SILENTLY DROPPED: ['hash']
```

This is latent rather than live: all 13 `KV_SERIALIZE*` macros currently defined in `contrib/epee/include/serialization/keyvalue_serialization.h` are covered by the alternation, so it cannot bite today. It only becomes reachable if epee gains a macro, which has happened upstream before.

The fix mirrors the guard you already wrote for routing — count the macro occurrences in each KV block and compare against the number parsed:

```python
total = len(re.findall(r"\bKV_SERIALIZE\w*\s*\(", struct_body))
if total != len(entries):
    raise ParseError(f"{total} KV_SERIALIZE macros in {name} but only {len(entries)} parsed")
```

Happy for this to land as a follow-up.

sn1f3rt (approved):
Verified against the branch: the generator reproduces the committed `docs/openapi.yaml` byte-for-byte, the spec passes `openapi-spec-validator` for OpenAPI 3.0, and every point from the previous review is addressed.

Two independent cross-checks I ran against the C++ source rather than against the generator's own output:

- Parsed `core_rpc_server_commands_defs.h` separately (following `KV_SERIALIZE_PARENT`, handling the `_N` rename forms) and diffed field sets against the spec across **132 command request/response structs — zero mismatches**.
- Cross-checked the `restricted` and `binary` flags for all **43 REST endpoints** against the routing table — **zero mismatches**.

I also confirmed the 16 operations without a `requestBody` are genuinely parameterless (`rpc_request_base` has an empty KV map, and Nerva's `COMMAND_RPC_UPDATE` really does have no `command` field, unlike Monero's).

The fail-closed guards go well past what I asked for. I re-ran the exact break scenario from the last review plus two others:

| Mutation | Result |
| --- | --- |
| `MAP_URI_AUTO_JON2` -> `MAP_URI_AUTO_JSON2` | `exit=1` `routing table parse mismatch: 84 MAP_* macros in core_rpc_server.h but only 50 were parsed` |
| New endpoint with no `HANDLER_DESCRIPTIONS` entry | `exit=1` `no description written for handler on_get_widget` |
| Field typed `boost::multiprecision::uint128_t` | `exit=1` `unmapped C++ type ... for field 'widget' in COMMAND_RPC_GET_HEIGHT (response)` |

Approving. One non-blocking gap noted inline, plus a maintenance note.

**Maintenance note:** the description guard means any newly registered RPC endpoint turns this CI job red until someone adds a `HANDLER_DESCRIPTIONS` entry. That is a reasonable trade for keeping the spec honest, but it is a real tax on future RPC changes and worth being a deliberate choice rather than a surprise.

**Heads-up, not a review item:** this branch is based on `f03145b`, two commits behind `master`. Neither touches `src/rpc`, so there is no drift today. Separately, the fix in #131 changes `/get_transaction_pool_hashes.bin` to `MAP_URI_AUTO_BIN2`, so whichever of the two lands second will need `docs/openapi.yaml` regenerated. The spec currently documents that endpoint as `application/json`, which is correct for this base.

XelisVault:
Now everything is ok !

sn1f3rt (approved):

PR #132, opened 2026-08-17
fix: serve /get_transaction_pool_hashes.bin as binary

### LINKED ISSUE #131: bug: /get_transaction_pool_hashes.bin is registered as JSON but returns a raw binary blob
`/get_transaction_pool_hashes.bin` is registered as a JSON endpoint but its response body is a raw binary blob, so the daemon emits a response that is neither valid JSON nor valid UTF-8.

### Details

`COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN::response_t` serializes its hash vector as a packed byte blob:

```cpp
struct response_t: public rpc_response_base
{
  std::vector<crypto::hash> tx_hashes;

  BEGIN_KV_SERIALIZE_MAP()
    KV_SERIALIZE_PARENT(rpc_response_base)
    KV_SERIALIZE_CONTAINER_POD_AS_BLOB(tx_hashes)
  END_KV_SERIALIZE_MAP()
};
```

`KV_SERIALIZE_CONTAINER_POD_AS_BLOB` stores the container as a `std::string` of raw bytes — `32 * n` bytes of hash material with no encoding applied.

The endpoint is nevertheless registered with the JSON macro at `src/rpc/core_rpc_server.h:122`:

```cpp
MAP_URI_AUTO_JON2("/get_transaction_pool_hashes.bin", on_get_transaction_pool_hashes_bin, COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN)
```

`MAP_URI_AUTO_JON2` routes through `load_t_from_json` / `store_t_to_json` (`contrib/epee/include/net/http_server_handlers_map2.h:61-87`), whereas `MAP_URI_AUTO_BIN2` uses the binary serializer. So the blob is emitted inside a JSON string literal, escaped by `transform_to_escape_sequence` (`contrib/epee/include/storages/parserse_base_utils.h:100-103`), which only escapes a fixed set:

```cpp
static const char escaped[] = "\b\f\n\r\t\v\"\\/";
```

Every other byte is passed through verbatim. Since tx hashes are effectively random, essentially every non-empty response contains unescaped control characters in the `0x00`-`0x1F` range and bytes `>= 0x80`. Unescaped `U+0000`-`U+001F` inside a string is a hard violation of RFC 8259, and the high bytes are not valid UTF-8. (Separately, the `\v` escape that function emits is not a JSON escape sequence either.)

### Impact

Any third-party consumer using a standards-compliant JSON parser — block explorers, pool software, monitoring, the OpenAPI/SDK tooling proposed in #130 — cannot parse this endpoint's response. Nerva's own wallet is unaffected because it calls the endpoint with `invoke_http_json` (`src/wallet/wallet2.cpp:2895`) and both ends therefore use epee's lenient parser, which round-trips the raw bytes; the breakage is only visible to external clients.

The bootstrap-daemon path is internally consistent with the JSON registration (`invoke_http_mode::JON` at `src/rpc/core_rpc_server.cpp:1548`), so this is not a crash or a sync failure — it is a wire-format correctness problem confined to external consumers.

### Divergence from upstream

Monero registers the `.bin` variant as binary and only the suffix-less variant as JSON:

```cpp
MAP_URI_AUTO_BIN2("/get_transaction_pool_hashes.bin", on_get_transaction_pool_hashes_bin, COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN)
MAP_URI_AUTO_JON2("/get_transaction_pool_hashes",     on_get_transaction_pool_hashes,     COMMAND_RPC_GET_TRANSACTION_POOL_HASHES)
```

Nerva uses `MAP_URI_AUTO_JON2` for both. This dates back to the initial rebases rather than a deliberate change: `5afce3b` ("initial rebase commit", 2019-04-21) carried the `.bin` path on the JSON macro, and `7b8da67` ("Initial 0.15 rebase commit", 2019-12-17) set the wallet caller to `invoke_http_json`.

### Affected files

- `src/rpc/core_rpc_server.h:122`
- `src/rpc/core_rpc_server.cpp:1544-1560`
- `src/wallet/wallet2.cpp:2895`

### Scope

Changing the registration to `MAP_URI_AUTO_BIN2` is a wire-format change on this endpoint and would need the wallet caller moved to the binary invocation in the same change, so daemon and wallet must be upgraded together. Worth deciding whether to match upstream or to keep the JSON path and change the response to a hex-encoded string array.

Found while reviewing #130, which would otherwise generate an OpenAPI spec documenting this endpoint as `application/json`.

sn1f3rt:
Fixes #131.

`/get_transaction_pool_hashes.bin` was registered with `MAP_URI_AUTO_JON2` even though `COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN::response_t` serialises its hash vector with `KV_SERIALIZE_CONTAINER_POD_AS_BLOB`. The result was `32 * n` bytes of raw hash material emitted inside a JSON string literal, escaped only by epee's `transform_to_escape_sequence`, which handles a fixed set of characters and passes everything else through verbatim. Since tx hashes are effectively random, practically every non-empty response carried unescaped `U+0000`-`U+001F` (an RFC 8259 violation) and bytes `>= 0x80` (not valid UTF-8).

## Changes

**Endpoint (`74b262c`)**

- `src/rpc/core_rpc_server.h` — register the path with `MAP_URI_AUTO_BIN2`, so it goes through `load_t_from_binary` / `store_t_to_binary` and is served as `application/octet-stream`.
- `src/rpc/core_rpc_server.cpp` — forward to the bootstrap daemon with `invoke_http_mode::BIN` to match.
- `src/wallet/wallet2.cpp` — `update_pool_state` now calls the endpoint with `invoke_http_bin`.
- `src/rpc/core_rpc_server_commands_defs.h` — `CORE_RPC_VERSION` 3.1 -> 4.0.

External consumers that want JSON should use the suffix-less `/get_transaction_pool_hashes`, which already exists and returns hex strings. That endpoint is unchanged.

**wallet-rpc version check (`926cf1a`)**

`simplewallet.cpp:4309` and `src/wallet/api/wallet.cpp:2032` both refuse a daemon whose RPC major version differs from their own. `wallet_rpc_server` had no equivalent check, so against a mismatched daemon it surfaced `get_tx_pool_error` with nothing naming the version. The major bump above makes that skew reachable for anyone pairing a wallet-rpc build with a separately-updated remote node, so the check lands here rather than separately.

- Startup path (`--wallet-file`, `--generate-from-json`): checked before the initial `wal->refresh()`, which is what would otherwise throw first.
- `--wallet-dir` path: checked in the open, create and restore handlers, before the new wallet is attached.
- An unreachable, offline or busy daemon reports no version and is not treated as a mismatch, so offline operation is unaffected.
- `--allow-mismatched-daemon-version` overrides, matching simplewallet.

The affected handlers now null `m_wallet` after deleting the outgoing wallet, so bailing out on a mismatch cannot leave a dangling pointer behind.

## Note on the "divergence from upstream" in the issue

The issue states that Monero registers the `.bin` variant as binary. That is not accurate — Monero v0.18.4.2 and current master both use `MAP_URI_AUTO_JON2` for `/get_transaction_pool_hashes.bin` and call it from `wallet2.cpp` with `invoke_http_json`, exactly as Nerva did. This is an inherited upstream quirk rather than a rebase artifact, so the change here is a deliberate divergence, not a realignment. The underlying wire-format bug is real either way, and the `.bin` suffix should mean binary.

## Compatibility

This is a wire-format change on the endpoint, so daemon and wallet must be upgraded together. A new wallet talking to a pre-4.0 daemon sends a binary body to a JSON-registered handler and gets a parse failure; the reverse fails the same way. Every wallet client now names the mismatch instead of failing obscurely.

Operationally, a 4.0 daemon running with `--bootstrap-daemon-address auto` will fail each pool forward to a pre-4.0 public node and rotate to another (`bootstrap_daemon::handle_result` disconnects when a next-node provider is set). This self-heals once the public fleet is on 4.0, but it is worth a release note, along with sequencing the public node upgrades against the CLI build that downstream wallets pull.

No in-tree consumers other than `wallet2.cpp` call the endpoint.

## Out of scope

The invalid `\v` escape emitted by `transform_to_escape_sequence` (`contrib/epee/include/storages/parserse_base_utils.h`) is also noted in the issue. It affects every JSON string the daemon and wallet emit, not just this endpoint, so it is left for a separate change.

## Testing

`obj_rpc`, `obj_daemon_rpc_server`, `obj_wallet` and `wallet_rpc_server` build clean on macOS arm64 (clang, Boost 1.89); only pre-existing warnings. `--allow-mismatched-daemon-version` verified present in `nerva-wallet-rpc --help`. The binary round-trip and the mismatch paths have not been exercised against a live daemon pair.

### DIFF
--- a/src/rpc/core_rpc_server.cpp
+++ b/src/rpc/core_rpc_server.cpp
@@ -1545,7 +1545,7 @@ namespace cryptonote
   {
     RPC_TRACKER(get_transaction_pool_hashes);
     bool r;
-    if (use_bootstrap_daemon_if_necessary<COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN>(invoke_http_mode::JON, "/get_transaction_pool_hashes.bin", req, res, r))
+    if (use_bootstrap_daemon_if_necessary<COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN>(invoke_http_mode::BIN, "/get_transaction_pool_hashes.bin", req, res, r))
       return r;
 
     const bool restricted = m_restricted && ctx;
--- a/src/rpc/core_rpc_server.h
+++ b/src/rpc/core_rpc_server.h
@@ -119,7 +119,7 @@ namespace cryptonote
       MAP_URI_AUTO_JON2_IF("/set_log_level", on_set_log_level, COMMAND_RPC_SET_LOG_LEVEL, !m_restricted)
       MAP_URI_AUTO_JON2_IF("/set_log_categories", on_set_log_categories, COMMAND_RPC_SET_LOG_CATEGORIES, !m_restricted)
       MAP_URI_AUTO_JON2("/get_transaction_pool", on_get_transaction_pool, COMMAND_RPC_GET_TRANSACTION_POOL)
-      MAP_URI_AUTO_JON2("/get_transaction_pool_hashes.bin", on_get_transaction_pool_hashes_bin, COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN)
+      MAP_URI_AUTO_BIN2("/get_transaction_pool_hashes.bin", on_get_transaction_pool_hashes_bin, COMMAND_RPC_GET_TRANSACTION_POOL_HASHES_BIN)
       MAP_URI_AUTO_JON2("/get_transaction_pool_hashes", on_get_transaction_pool_hashes, COMMAND_RPC_GET_TRANSACTION_POOL_HASHES)
       MAP_URI_AUTO_JON2("/get_transaction_pool_stats", on_get_transaction_pool_stats, COMMAND_RPC_GET_TRANSACTION_POOL_STATS)
       MAP_URI_AUTO_JON2_IF("/set_bootstrap_daemon", on_set_bootstrap_daemon, COMMAND_RPC_SET_BOOTSTRAP_DAEMON, !m_restricted)
--- a/src/rpc/core_rpc_server_commands_defs.h
+++ b/src/rpc/core_rpc_server_commands_defs.h
@@ -86,8 +86,8 @@ namespace cryptonote
 // whether they can talk to a given daemon without having to know in
 // advance which version they will stop working with
 // Don't go over 32767 for any of these
-#define CORE_RPC_VERSION_MAJOR 3
-#define CORE_RPC_VERSION_MINOR 1
+#define CORE_RPC_VERSION_MAJOR 4
+#define CORE_RPC_VERSION_MINOR 0
 #define MAKE_CORE_RPC_VERSION(major,minor) (((major)<<16)|(minor))
 #define CORE_RPC_VERSION MAKE_CORE_RPC_VERSION(CORE_RPC_VERSION_MAJOR, CORE_RPC_VERSION_MINOR)
 
--- a/src/wallet/wallet2.cpp
+++ b/src/wallet/wallet2.cpp
@@ -2892,7 +2892,7 @@ void wallet2::update_pool_state(std::vector<std::tuple<cryptonote::transaction, 
 
   {
     const boost::lock_guard<boost::recursive_mutex> lock{m_daemon_rpc_mutex};
-    bool r = epee::net_utils::invoke_http_json("/get_transaction_pool_hashes.bin", req, res, m_http_client, rpc_timeout);
+    bool r = epee::net_utils::invoke_http_bin("/get_transaction_pool_hashes.bin", req, res, m_http_client, rpc_timeout);
     THROW_ON_RPC_RESPONSE_ERROR(r, {}, res, "get_transaction_pool_hashes.bin", error::get_tx_pool_error);
   }
   MTRACE("update_pool_state got pool");
--- a/src/wallet/wallet_rpc_server.cpp
+++ b/src/wallet/wallet_rpc_server.cpp
@@ -70,6 +70,26 @@ namespace
   const command_line::arg_descriptor<bool> arg_restricted = {"restricted-rpc", "Restricts to view-only commands", false};
   const command_line::arg_descriptor<std::string> arg_wallet_dir = {"wallet-dir", "Directory for newly created wallets"};
   const command_line::arg_descriptor<bool> arg_prompt_for_password = {"prompt-for-password", "Prompts for password when not provided", false};
+  const command_line::arg_descriptor<bool> arg_allow_mismatched_daemon_version = {"allow-mismatched-daemon-version", "Allow communicating with a daemon that uses a different RPC version", false};
+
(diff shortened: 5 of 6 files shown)


ngeojiajun (approved):

R0BC0D3R:
Reviewed this. The change is correct and I found nothing that breaks the daemon or consensus.

**Premise checks out.** `transform_to_escape_sequence` escapes only `\b \f \n \r \t \v " \ /` and passes every other byte through, so a 32-byte hash produces invalid JSON with ~98.6% probability (`1 - (7/8)^32` for at least one byte in `0x00`-`0x1F`) and invalid UTF-8 with probability ~1. It also emits `\v`, which is not a JSON escape at all.

**Worth adding to the rationale:** this was the only `.bin` path in `core_rpc_server.h` registered as JSON. The other nine already use `MAP_URI_AUTO_BIN2`, and `wallet2.cpp` already calls `invoke_http_bin` in eight other places, so this makes the outlier consistent rather than introducing something new.

Also confirmed: the request struct is empty, `use_bootstrap_daemon_if_necessary` genuinely dispatches `invoke_http_mode::BIN`, `wallet2.cpp:2895` is the only in-tree consumer, and the version bump is safe because every other comparison is against `MAKE_CORE_RPC_VERSION(1, x)`, which 4.0 clears exactly as 3.1 did.

Two things worth deciding before merge:

1. **`wallet_rpc_server` has no RPC major-version check.** `simplewallet.cpp:4309` and `wallet/api/wallet.cpp:2032` both have one; wallet-rpc has none. So a 4.0 wallet-rpc against a 3.1 daemon starts cleanly and then throws `get_tx_pool_error` on every refresh with nothing pointing at the version.

   That has a concrete consumer here. NervaOne's Wallet Only mode runs the downloaded `nerva-wallet-rpc.exe` against a user-supplied remote node, so the two ends are not from the same build and skew is a normal outcome. Full Node and Pruned Node ship both binaries in one archive and are never affected. Since wallet-rpc is the only client that reports a mismatch as an opaque pool error rather than naming it, worth landing the check in this PR rather than deferring it.

2. **Bootstrap forwarding churns against un-upgraded peers.** `bootstrap_daemon::invoke_http_bin` wraps the call in `handle_result`, which disconnects when `m_get_next_public_node` is set. With `--bootstrap-daemon-address auto`, every pool query then rotates to another public node that also cannot answer, for the whole transition window. It self-heals once the fleet is on 4.0, so I would accept it, but it belongs in the release notes.

Minor: worth a comment on the registration line pointing at #131. Upstream still uses `MAP_URI_AUTO_JON2` here, so this is a deliberate divergence that a future rebase could silently revert.

Deployment note: this is a flag-day change in both directions. Sequence the public node upgrades with the CLI build NervaOne downloads, so Wallet Only users are not stranded between versions, and land the wallet-rpc check first so the one client without a version check is not the one left guessing.

PR #133, opened 2026-08-28
fix(db): block removal left stale entries in the proof of work seed cache

R0BC0D3R:
## Problem

`m_block_cache` is a height-indexed cache of block data that feeds the CryptoNight-Adaptive proof of work seed (`get_cna_v2_data`, `get_cna_v6_data`). Nothing invalidated it when blocks were removed: only `close()` and `reset()` cleared it, and `build_block_cache` only ever appended.

There are three ways the chain can lose blocks and two of them left the cache stale. `remove_block()` covers every pop and every reorg. `batch_abort()` is the less obvious one: `build_block_cache` reads through the batch write transaction by design, so the cache legitimately covers blocks that are still uncommitted, and an abort discards them without `remove_block()` ever running.

In both cases `m_block_cache_height` kept claiming heights whose contents had changed, so later seeds were drawn from a chain that no longer existed. The node then computed a different long hash than the network for every block and rejected each as bad proof of work, while still looking connected.

## When it bites

Live on v0.3.0.0 today, through the recovery path. After a `pop_blocks` deeper than the cache frontier, the entries above the new tip survive and `build_block_cache` short-circuits on `have >= height`, so the node keeps seeding from records the pop removed. **This is why popping alone has never recovered a forked node and a restart has always been needed.**

Observed on mainnet 2026-08-31: a node stalled at 4377750, was popped 500 blocks, resynced, stalled again at the same height with the identical computed hash, and only recovered once a restart rebuilt the cache from disk.

Separately, `warm_block_cache` inside `ensure_batch_longhashes` (added by #117) lifts the cache to the tip during catch-up sync. That widens the exposure from deep pops to ordinary reorgs once HF14 activates at 4,500,000.

Complementary to #136, which fixes a separate on-disk corruption. Recovering an already-forked node needs both: the pop writes correct difficulties back to disk, and this stops the cache from serving the stale ones.

## Fix

- `remove_block()` lowers `m_block_cache_height`. Single chokepoint for pops and reorgs.
- `batch_abort()` lowers it to what actually committed.
- `build_block_cache()` refills from that height and overwrites the stale slots, and re-checks the height under the lock so a concurrent filler's entries are not discarded.

The height is lowered but the vector is never shrunk. Readers resolve a height before taking the shared lock, so shrinking could drop the buffer under a live reader and walk it off the end (same class as #108). Leaving the entries in place and overwriting them on the way back up is equally correct and cannot fault.

Not a consensus change. The rule is unchanged, this only stops a stale cache from breaking it. No fork height, no reindex, and safe to deploy node by node, since the stale state was always node-local.

## Testing

Two tests that look right and prove nothing: a forward sync never calls `remove_block()`, and `pop_blocks` plus a resync re-downloads the same blocks, so the stale entries match and it passes on master too. It needs a real divergence, and the cache has to reach above the fork point.

Two isolated daemons at a common height `F`:

1. Miner mines 100 blocks. The test node syncs them in one batch, which is what runs `warm_block_cache` and lifts the cache to the tip. Confirm with a `Hashed N blocks on N threads` line (needs `--show-time-stats 1`); without it the cache never rises above `tip - 255` and the test is vacuous.
2. Isolate. The miner pops back to `F` and mines a longer chain reaching at least `F + 260`.
3. Reconnect and let the test node reorg. Do not restart it, since a restart rebuilds the cache from disk and hides the bug.

Keep the first chain under 256 blocks so the reorg can complete at all, and make the second exceed `F + 257`, which is where the first poisoned slot is read.

Expected: on master the test node reorgs and then stalls at `F + 257`. On this branch it follows to the tip. Run both arms, since the failure on master is what proves the test reaches the bug.

### DIFF
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -800,6 +800,22 @@ void BlockchainLMDB::remove_block()
 
   if ((result = mdb_cursor_del(m_cur_block_info, 0)))
       throw1(DB_ERROR(lmdb_error("Failed to add removal of block info to db transaction: ", result).c_str()));
+
+  // m_block_cache feeds the proof of work seed by height (get_cna_v2_data,
+  // get_cna_v6_data) and nothing else invalidated it. A popped block left in it
+  // makes later seeds draw on a chain that no longer exists, so the node hashes
+  // differently from the network and rejects every block as bad proof of work.
+  // Seeds read 256 blocks back so the cache normally trails the tip;
+  // ensure_batch_longhashes warms it to the tip during catch-up sync.
+  //
+  // Lower the height, keep the vector: readers resolve a height before taking
+  // the shared lock, so shrinking could walk one off the end.
+  {
+    const uint64_t new_height = m_height - 1;
+    boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
+    if (m_block_cache_height.load(std::memory_order_relaxed) > new_height)
+      m_block_cache_height.store(new_height, std::memory_order_release);
+  }
 }
 
 uint64_t BlockchainLMDB::add_transaction_data(const crypto::hash& blk_hash, const std::pair<transaction, blobdata>& txp, const crypto::hash& tx_hash, const crypto::hash& tx_prunable_hash)
@@ -2395,6 +2411,12 @@ void BlockchainLMDB::build_block_cache(uint64_t height)
 
   boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
 
+  // Re-check under the lock: another thread may have filled the cache while we
+  // waited, and storing our own smaller height below would drop its entries.
+  const uint64_t have = m_block_cache_height.load(std::memory_order_relaxed);
+  if (have >= height)
+    return;
+
   m_block_cache.reserve(height);
 
   // Read block_info via the standard batch-aware cursor pattern. During batch
@@ -2410,7 +2432,9 @@ void BlockchainLMDB::build_block_cache(uint64_t height)
   TXN_PREFIX_RDONLY();
   RCURSOR(block_info);
 
-  for (uint64_t index = m_block_cache.size(); index < height; ++index)
+  // Start from the cached height, not the vector size: remove_block lowers the
+  // height without shrinking, so these slots may hold popped blocks. Overwrite.
+  for (uint64_t index = have; index < height; ++index)
   {
     MDB_val_set(query, index);
     if (auto r = mdb_cursor_get(m_cur_block_info, (MDB_val*)&zerokval, &query, MDB_GET_BOTH))
@@ -2421,7 +2445,10 @@ void BlockchainLMDB::build_block_cache(uint64_t height)
     entry.timestamp = bi->bi_timestamp;
     entry.diff_lo   = bi->bi_diff_lo;
     entry.coins     = bi->bi_coins;
-    m_block_cache.push_back(entry);
+    if (index < m_block_cache.size())
+      m_block_cache[static_cast<size_t>(index)] = entry;
+    else
+      m_block_cache.push_back(entry);
   }
 
   m_block_cache_height.store(height, std::memory_order_release);
--- a/src/blockchain_db/lmdb/db_lmdb.cpp
+++ b/src/blockchain_db/lmdb/db_lmdb.cpp
@@ -802,12 +802,8 @@ void BlockchainLMDB::remove_block()
       throw1(DB_ERROR(lmdb_error("Failed to add removal of block info to db transaction: ", result).c_str()));
 
   // m_block_cache feeds the proof of work seed by height (get_cna_v2_data,
-  // get_cna_v6_data) and nothing else invalidated it. A popped block left in it
-  // makes later seeds draw on a chain that no longer exists, so the node hashes
-  // differently from the network and rejects every block as bad proof of work.
-  // Seeds read 256 blocks back so the cache normally trails the tip;
-  // ensure_batch_longhashes warms it to the tip during catch-up sync.
-  //
+  // get_cna_v6_data). A popped block left in it makes later seeds read a chain
+  // that no longer exists, and every block after that fails as bad PoW.
   // Lower the height, keep the vector: readers resolve a height before taking
   // the shared lock, so shrinking could walk one off the end.
   {
@@ -2411,8 +2407,8 @@ void BlockchainLMDB::build_block_cache(uint64_t height)
 
   boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
 
-  // Re-check under the lock: another thread may have filled the cache while we
-  // waited, and storing our own smaller height below would drop its entries.
+  // Re-check under the lock: another thread may have filled it while we waited,
+  // and storing our own smaller height below would discard its entries.
   const uint64_t have = m_block_cache_height.load(std::memory_order_relaxed);
   if (have >= height)
     return;
@@ -2433,7 +2429,7 @@ void BlockchainLMDB::build_block_cache(uint64_t height)
   RCURSOR(block_info);
 
   // Start from the cached height, not the vector size: remove_block lowers the
-  // height without shrinking, so these slots may hold popped blocks. Overwrite.
+  // height without shrinking, so these slots may hold popped blocks.
   for (uint64_t index = have; index < height; ++index)
   {
     MDB_val_set(query, index);
@@ -4087,6 +4083,18 @@ void BlockchainLMDB::batch_abort()
   m_write_batch_txn = nullptr;
   m_batch_active = false;
   memset(&m_wcursors, 0, sizeof(m_wcursors));
+
+  // build_block_cache reads through the batch write txn, so the cache can cover
+  // blocks this abort just discarded, and remove_block never ran for them. Same
+  // stale seed data a reorg would leave. height() after the abort reflects what
+  // actually committed.
+  {
+    const uint64_t committed = height();
+    boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
+    if (m_block_cache_height.load(std::memory_order_relaxed) > committed)
+      m_block_cache_height.store(committed, std::memory_order_release);
+  }
+
   LOG_PRINT_L3("batch transaction: aborted");
 }
 


src/blockchain_db/lmdb/db_lmdb.cpp
```diff
@@ -800,6 +800,22 @@ void BlockchainLMDB::remove_block()
...
+  // ensure_batch_longhashes warms it to the tip during catch-up sync.
+  //
+  // Lower the height, keep the vector: readers resolve a height before taking
+  // the shared lock, so shrinking could walk one off the end.
+  {
+    const uint64_t new_height = m_height - 1;
+    boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
+    if (m_block_cache_height.load(std::memory_order_relaxed) > new_height)
+      m_block_cache_height.store(new_height, std::memory_order_release);
+  }
```
sn1f3rt, line 814: `remove_block()` is the chokepoint for *pops*, but not for every path that unmakes a block: a batch abort discards blocks without ever calling it, and leaves the cache in exactly the state this hunk is fixing.

`ensure_batch_longhashes` warms the cache on the batch-owning thread (`blockchain.cpp:4462`), and `build_block_cache` reads through the batch write txn, so the cache legitimately covers blocks that are still uncommitted. If the batch then aborts -- `m_batch_success = false` at `blockchain.cpp:3762`/`3771`, then `batch_abort()` at `blockchain.cpp:4110` -- the db reverts to the batch start height but `m_block_cache_height` still points past it, holding the rolled-back blocks. `BlockchainLMDB::batch_abort()` touches only the txn and cursors.

Same failure mode as the one fixed here, and arguably easier to hit than a reorg: the node re-syncs those heights, and if it lands on a different chain than the batch it just threw away (the likely reason the batch failed), the seeds 256 blocks later read blocks that never existed.

Symmetric fix, same shape as this hunk, in `batch_abort()` after `m_batch_active = false`:

```cpp
{
  const uint64_t committed = height();
  boost::unique_lock<boost::shared_mutex> cache_lock(m_block_cache_lock);
  if (m_block_cache_height.load(std::memory_order_relaxed) > committed)
    m_block_cache_height.store(committed, std::memory_order_release);
}
```

The `pop_block` failure path is fine as-is -- `block_wtxn_abort()` rolls back the delete while the height stays lowered, which is the safe direction.

R0BC0D3R: Confirmed and applied in 8453234.

You're right that `remove_block()` isn't the chokepoint I claimed. Traced the path you describe: `batch_abort()` touches only the txn, cursors and flags, `m_batch_success = false` is set at `blockchain.cpp:3762`/`:3771`, and `cleanup_handle_incoming_blocks` reaches `batch_abort()` at `:4110`. Since `build_block_cache` reads through the batch write txn by design, the cache really can cover blocks the abort discards, and nothing lowers the height for them.

Your point that it's easier to hit than a reorg holds too. The batch aborts *because* something went wrong, and the node then re-syncs those heights, which is exactly when it can land on a different chain than the one it just threw away.

Took your snippet as-is, placed at the end of `batch_abort()` after `m_batch_active = false` and the cursor memset, so `height()` opens a fresh read-only txn and returns what actually committed, and the cache lock isn't held across the db read.

Agreed on the `pop_block` failure path. `block_wtxn_abort()` restores the block while the height stays lowered, and lowering is the safe direction since the cache just refills.

Also rewrote the testing section in the description. The recipe there couldn't work: it called for a fork deeper than 256, which can never reorg at all, because alt blocks that far above the split read their seed from the main chain and fail PoW for an unrelated reason. It would have passed identically on master. Replaced with a sub-256 fork that relies on `warm_block_cache` to lift the cache above the fork point.

sn1f3rt:
Traced the mechanism for the "What I have not established" part. It is
`Blockchain::get_difficulty_for_next_block`, and you were right that
`handle_alternative_block`, `switch_to_alternative_blockchain` and
`pop_block_from_blockchain` are all correct. They set
`m_timestamps_and_difficulties_height = 0` as they should. The problem is that
something sets it straight back.

That function is called from RPC threads with no blockchain lock: `on_get_info`
(`core_rpc_server.cpp:328`), `on_mining_status` (`:1367`), zmq `get_info`. Those
threads are not the batch writer, so `block_rtxn_start` gives them an
`MDB_RDONLY` txn on the last committed state, and `prepare_handle_incoming_blocks`
releases the blockchain lock on return while its batch stays open across the
whole per-block loop. So an RPC poll can hold the blockchain lock, mid-batch,
and read the chain you just reorged away.

It rebuilds `m_timestamps`/`m_difficulties` from that view and republishes
`m_timestamps_and_difficulties_height`, which is keyed by height alone. A reorg
keeps the height and changes the contents. The writer's next call then sees a
cached height exactly one below its own, takes the incremental path, and appends
an accepted block to an orphaned window:

```
committed: ... 4377488, 4377489-A          (orphan tip, height 4377490)
in batch:  ... 4377488, 4377489-B, 4377490-B

writer  height 4377491, window ends 4377490-B      correct
rpc     height 4377490, rebuilds from chain A      m_t_a_d_height = 4377490
        batch commits
writer  height 4377491, delta == 1 -> incremental
        window = [4377430..4377490] with 4377489-A at index 59
```

Index 59, the orphan's timestamp, two seconds. Your 2492059, to the digit.

I reproduced your arithmetic independently before trusting it. Your LWMA times
k=1830 is exactly 108777, so it is consistent with the real `next_difficulty_v6`;
canonical gives 2492013 and the +2s perturbation gives 2492059.03 -> 2492059. And
your point about #120 holds quantitatively: the canonical result sits 0.21 from
the integer boundary, relative 8.4e-8, nine orders above double rounding.

This also explains the shape of the reports. Both boxes run the GUI, which polls
`get_info` and `mining_status` about once a second, and the reorg batch is open
for as long as it takes to verify two CNA v6 blocks. That is a wide window, hit
several times a day on a seed node, and it is OS independent, which matches the
Windows VPS going too.

Filed as #135 with the fix in #136. It keys the window by the hash of its newest
block, the way the difficulty value cache at `:867` already is.

Kept separate from this PR deliberately: no file overlap, and they are different
bugs. Yours is an in-memory cache that a restart clears; this one writes wrong
consensus data to disk that a restart does not. But they want to ship in the same
release, because recovering an already-forked node needs both, exactly as you saw
on your VPS: popping blocks writes correct difficulties back to disk, and without
this PR `m_block_cache` keeps serving the stale `diff_lo`.

One note for the block cache while you are in there. `build_block_cache` has the
same non-writer-stale-view shape, but `ensure_batch_longhashes` sidesteps it by
warming on the batch-owning thread so workers never open a cursor. The one path
still exposed is the unrestricted `calc_pow` RPC landing after a reorg has
lowered the cache height. Narrow, and not a blocker for this PR.

Also agree the description's "latent until HF14" line should change. The stall
you produced by popping and restarting is on v0.3.0.0 today.

R0BC0D3R:
Thanks for tracing it, and the GUI polling detail is the piece I was missing. Once you say the boxes poll `get_info` and `mining_status` about once a second against a batch that stays open for two CNA v6 verifications, the frequency and the OS independence both fall out. I had the arithmetic pointing at index 59 but no mechanism to put an orphan there.

Reviewed #136 in detail over on that PR. Agreed on all of it, including keeping the two separate and shipping them together.

Two corrections to the closing notes.

**`calc_pow` does not exist in this fork.** No RPC, no command, no `COMMAND_RPC_CALC_POW` anywhere in `src/`, and it is absent from the JSON-RPC method map. That endpoint is upstream Monero only, added after this fork's base, so that particular path is not reachable here.

The shape you are pointing at is real though, it just lands on a different door. The only non-writer caller of `get_block_longhash` in the tree is `miner.cpp:870`; no RPC path reaches the long hash at all. A mining thread gets an `MDB_RDONLY` txn on committed state, so mid-batch it could refill `build_block_cache` from a stale view and raise the height back after `remove_block` lowered it.

Narrower than the difficulty case and not a factor in these incidents:

- mining nodes only, and both seed nodes run `not mining`
- `miner::pause()` only increments `m_pausers_count` and returns without joining, so a worker already inside `get_block_longhash` keeps going, and a CNA v6 hash is ~200ms, so the window is small but not zero

Not a blocker for this PR. It does argue for eventually keying the block cache by content the way #136 keys the difficulty window, rather than by height plus invalidation hooks. Happy to open that as a follow-up issue if you agree it is worth tracking.

**Description updated.** Dropped the latent-until-HF14 line. It now leads with the recovery path being live on v0.3.0.0, with the mainnet sequence from last night: stalled at 4377750, popped 500, resynced, stalled again at the same height with the identical computed hash, recovered only after a restart. `warm_block_cache` is now framed as what widens it from deep pops to ordinary reorgs at HF14, rather than as the thing that makes it live at all.

sn1f3rt:
You are right about `calc_pow` and I was wrong. `grep -rni "calc_pow\|calcpow" src/` returns nothing, there is no `COMMAND_RPC_CALC_POW`, and it is not in the method map. I carried that one over from upstream without checking it existed here. Thanks for catching it.

Verified `miner::pause()` too, and it is as you describe: it takes `m_miners_count_lock`, increments `m_pausers_count`, logs, returns. No join, no wait for a worker to leave the hash. So a thread already inside `get_block_longhash` runs to completion.

One precision on the non-writer callers, because it changes what the follow-up is for. `miner.cpp:870` is not the only one. `Blockchain::longhash_worker` (`blockchain.cpp:4364`) also calls `get_block_longhash` and runs on threadpool threads, so those are non-writers too. They are safe, but only because `ensure_batch_longhashes` warms the cache on the batch-owning thread first, which is exactly what the comment at `:4468` says it is for. `build_block_cache` is then a no-op inside the workers.

So the accurate statement is that the miner is the only *unprotected* non-writer caller. The worker path is correct by an invariant rather than by construction: it holds as long as the warm call stays ahead of every height a worker will ask for. That is a reasonable invariant today and it is documented, but it is the kind that a later change to chunking or seed depth breaks silently, and the failure mode is the one we are already dealing with. Which is your argument for content-keying, and I agree with it. Please do open the follow-up.

One correction going the other way. On the exposure argument, the box in your first report was mining: the screenshot shows `Miner: Mining` at 61 h/s with 52 days uptime. The second Windows VPS was the idle one. So at least one of the forked nodes was mining at the time. It does not change the diagnosis, since the difficulty path accounts for 2492059 exactly and needs no help, but it means the miner path was not excluded by the conditions on the night, only by being narrower.

Also worth noting what your updated description now documents, because it is the joint repro for both PRs:

- stall at 4377750, which is 4377491 + 259, the first height whose seed window can reach the corrupted record
- pop 500 and resync writes correct difficulties back to disk, but stalls again at the same height with the identical hash, because `m_block_cache` never lowered and keeps serving the stale `diff_lo`
- restart rebuilds the cache from the now correct disk and it recovers

That is #136 writing the bad record and this PR keeping it alive after the disk was repaired. Neither one alone gets a forked node back. Good evidence for shipping them together, and worth keeping in the release notes as the recovery sequence.

R0BC0D3R:
Verified all of this against the branch.

**Your corrections hold.** `calc_pow` returns zero hits in `src/`. `miner::pause()` is exactly as you describe: lock, increment, log, return, no join. And you are right that `longhash_worker` is a second non-writer caller, so my "only" was wrong. The full set is `miner.cpp:870`, `blockchain.cpp:1739` and `:3559` (both writer-context), and `blockchain.cpp:4361` in `longhash_worker`. That call is at 4361, not 4364.

**One place the code is stronger than you credit.** The chunking half of that invariant is already defended: `ensure_batch_longhashes` clamps `threads = std::min(threads, CN_SEED_LOOKBACK_BLOCKS - 1)` with a comment saying why. Workers hash `blockchain_height + i` for `i <= 254`, whose seeds reach down to `blockchain_height - 2`, against a cache warmed to `blockchain_height`. Two blocks of headroom, and raising `prep-blocks-threads` cannot break it.

The seed-depth half is the real risk, and it is smaller than a rewrite. `CN_SEED_LOOKBACK_BLOCKS = 256` in `blockchain.cpp` is a hand-copied duplicate of the `height - 256` in `get_block_longhash_v13`/`_v14` in `cryptonote_tx_utils.cpp`, with nothing linking them. Change the seed depth there and this clamp silently goes wrong, with the failure mode we are already debugging. A shared constant or a `static_assert` covers it, and I will open the follow-up with that as the concrete item.

**The screenshot was misread, and I should have labelled it.** That box was the healthy one, sent to show 52 days uninterrupted. It reads `Your Height 4378619` equal to `Net Height`, every peer at the same height, `Sync OK`, and 4378619 is 869 blocks past the 4377750 stall. It went through the fork point and kept mining. So the forked boxes were the non-mining ones, and the miner path is excluded by the conditions on the night rather than merely narrower. The exposure is still real and still worth tracking, it just contributed nothing here.

**Hunks re-verified.** `remove_block`: `m_height` is the local captured at entry, before any deletion, so `m_height - 1` is correct and not off by one. `batch_abort`: abort, delete and null all complete before `height()`, so it reads committed state with no use-after-free, and the cache lock is not held across the db read. `build_block_cache`: acquire fast path plus relaxed re-check under the lock, and starting the loop at `have` rather than `m_block_cache.size()` is the load-bearing part, since `have <= size` always holds.

sn1f3rt (approved):

PR #136, opened 2026-08-31
fix(core): a stale reader poisoned the difficulty window cache

### LINKED ISSUE #135: fix: nodes fork off after a routine reorg writes a wrong difficulty to disk
## Summary

Nodes are forking off the network on their own, with no operator action and no
bad peer involved. Both seed nodes have done it repeatedly, on Linux and on
Windows, mining and not mining. The node stays connected, keeps its peer list,
reports a plausible height, and then stops following the chain.

The trigger is an ordinary 1-block orphan race, which happens on this network
several times a day.

## What actually happens

A routine reorg causes the node to write a wrong cumulative difficulty to disk
for one block. The error is small, a few units out of ~2.5M, so the block still
satisfies `check_hash` and is accepted rather than rejected. Nothing is logged.

Because the value stored per block is *cumulative*, every block after it
inherits the same offset. From that point the node's `block_info` records
disagree with the network's for every subsequent height.

Since HF13, `get_cna_v6_data` folds `bi_diff_lo` into the CryptoNight-Adaptive
proof of work seed, and the seed is taken 256 blocks back
(`stable_height = height - 256`). So 259 blocks after the corrupted record, the
node begins computing a different long hash than the network for essentially
every block, rejects each one as bad proof of work, and stalls.

## Observed instance

Height 4377489 was replaced by a 1-block reorg at 19:40:45. The accepted block's
timestamp is 1788118761; the orphan's is 1788118763, two seconds later.

Over the canonical window for block 4377491 (heights 4377430-4377490,
`DIFFICULTY_BLOCKS_COUNT_V6` = 61) the network's `next_difficulty_v6` gives:

```
LWMA           = 59.440983606557374
harmonic_mean  = 2473742.7582577863
nextDifficulty = 2492013.2100259392  ->  2492013
```

The node stored **2492059**. That is exactly the value produced by substituting
the orphan's timestamp at index 59 of the window, which is height 4377489:
shifting one interior timestamp by +2s nets `d sum(solveTime*i) = -2`, so
`LWMA = (108777-2)/1830` and `nextDifficulty = 2492059.03 -> 2492059`.

Floating point nondeterminism is ruled out: the canonical result lands 0.21 away
from the integer boundary, a relative distance of 8.4e-8, nine orders of
magnitude above double rounding. This is not the class of problem in #120.

A block mined against 2492013 satisfies a 2492059 target with probability
1 - 1.8e-5, which is why the block is accepted and the wrong value persists.

## Impact

- Silent consensus divergence from a routine event. No error until the node has
  already been on its own chain for 259 blocks.
- Not self-healing. The bad value is on disk, so restarting does not clear it.
  Recovery requires popping back below the first corrupted height and resyncing.
- Latent before HF13, because the stored difficulty did not feed the proof of
  work. Live on mainnet since block 4320000.
- Seed nodes are the most exposed, since more peers means more tip races.

## Affected

- `src/cryptonote_core/blockchain.cpp` - `Blockchain::get_difficulty_for_next_block`
- `src/blockchain_db/lmdb/db_lmdb.cpp` - `get_cna_v6_data` (amplifier, not cause)

## Scope

Node-local corruption written to node-local storage. The consensus rule is
unchanged and the network's chain is unaffected. Separate from #133, which
concerns the in-memory block cache and is not the cause of this.

sn1f3rt:
Closes #135.

## Cause

`get_difficulty_for_next_block` is called from RPC threads with no blockchain
lock: `on_get_info` (`core_rpc_server.cpp:328`), `on_mining_status` (`:1367`)
and the zmq `get_info` (`daemon_handler.cpp:442`). The comment above the fast
path says that is fine because the answer may just be slightly out of date. It
is not fine, for a reason unrelated to staleness.

Those threads are not the batch writer. `block_rtxn_start` returns the write txn
only to `m_writer`; everyone else gets an `MDB_RDONLY` txn on the last committed
state. `prepare_handle_incoming_blocks` scopes the blockchain lock to itself and
releases it on return, while the batch it opened stays open across the whole
per-block loop. So an RPC thread can hold the blockchain lock, mid-batch, and
read a chain the writer has already reorged away.

It then rebuilds `m_timestamps`/`m_difficulties` from that view and publishes
them under `m_timestamps_and_difficulties_height`, which is keyed by height
alone. A reorg keeps the height and changes the contents. Every reorg path
already resets the height to 0 (`:1016`, `:602`, `:620`), which is why an audit
of those resets comes up clean; the reader sets it straight back afterwards,
from a chain that no longer exists.

The writer's next call then finds a cached height exactly one below its own,
takes the incremental path, and appends an accepted block to an orphaned window:

```
committed: ... 4377488, 4377489-A          (orphan tip, height 4377490)
in batch:  ... 4377488, 4377489-B, 4377490-B

writer  height 4377491, window ends 4377490-B      correct
rpc     height 4377490, rebuilds from chain A      m_t_a_d_height = 4377490
        batch commits
writer  height 4377491, delta == 1 -> incremental
        window = [4377430..4377490] with 4377489-A
```

That is the observed 2492059 in #135, to the digit.

## Fix

Key the window by the hash of its newest block, exactly as the difficulty value
cache at `:867` already is. Identical hash at identical height implies identical
ancestry, so validating the newest entry validates the whole window. A mismatch
costs one full 61-block rebuild.

Also take a single read transaction for the tip read and the window. Each lookup
otherwise renews its own, so a 61-block window is 122 independent snapshots and
a commit landing mid-loop can splice two chains. This is hardening rather than
part of the fix, and it removes 121 renew/reset pairs per `get_info`.

Not a consensus change. The rule is unchanged; this stops a stale cache from
breaking it. No fork height, no reindex, safe node by node.

## Testing

The cheap tests prove nothing here: a forward sync never diverges the two views,
and a single-threaded reorg never has a reader in the gap. The bug needs a reorg
inside an open batch and a concurrent RPC poll.

**Transcription test.** I transcribed the cache logic with both LMDB views and
ran the orderings. Sequencing matters: the difficulty for block `h` is computed
while the chain height is `h`, and using the post-add view instead hides the
bug entirely.

| case | master | this branch |
| --- | --- | --- |
| reorg, no RPC poll (control) | clean | clean |
| reorg + RPC poll inside the batch | corrupt @4377489 | clean |
| repeated polls | corrupt | clean |
| 3-block reorg + poll | corrupt @4377487-89 | clean |
| poll straddling the commit, both orderings | clean | clean |
| batch_abort then re-add | clean | clean |

Fast path unaffected: 1 full rebuild per 200 blocks, identical to master.

**Arithmetic.** Independently reproduced #135's numbers. The reported LWMA times
k=1830 is exactly 108777, so it is consistent with the real `next_difficulty_v6`.
Canonical gives 2492013; one interior timestamp at +2s gives 2492059.03 ->
2492059. The canonical result sits 0.21 from the integer boundary, relative
8.4e-8, so #120's class of problem is excluded.

**Live repro**, if someone wants it end to end: two isolated daemons at a common
height, poll `get_info` in a loop at 1 Hz against the test node, then feed it a
span that reorgs its tip and continues for at least 260 blocks. On master it
stores a difficulty that disagrees with the miner's for the block after the
reorg and stalls 259 blocks later; on this branch it follows. The poll is the
part that makes it reproduce.

## Relationship to #133

Independent bugs, no file overlap, rebases cleanly either way. #133 is an
in-memory cache that a restart clears. This one writes wrong consensus data to
disk that a restart does not clear.

They should ship in the same release, because recovering an already-forked node
needs both: popping blocks writes correct difficulties back to disk, but without
#133 `m_block_cache` keeps serving the stale `diff_lo`.

### DIFF
--- a/src/cryptonote_core/blockchain.cpp
+++ b/src/cryptonote_core/blockchain.cpp
@@ -86,7 +86,7 @@ DISABLE_VS_WARNINGS(4267)
 
 //------------------------------------------------------------------
 Blockchain::Blockchain(tx_memory_pool& tx_pool) :
-  m_db(), m_tx_pool(tx_pool), m_hardfork(NULL), m_hash_context(NULL), m_timestamps_and_difficulties_height(0), m_current_block_cumul_weight_limit(0), m_current_block_cumul_weight_median(0),
+  m_db(), m_tx_pool(tx_pool), m_hardfork(NULL), m_hash_context(NULL), m_timestamps_and_difficulties_height(0), m_timestamps_and_difficulties_top_hash(crypto::null_hash), m_current_block_cumul_weight_limit(0), m_current_block_cumul_weight_median(0),
   m_enforce_dns_checkpoints(false), m_max_prepare_blocks_threads(4), m_batch_longhash_base(0), m_prepare_blocks(NULL), m_prepare_nblocks(0), m_db_sync_on_blocks(true), m_db_sync_threshold(1), m_db_sync_mode(db_async), m_db_default_sync(false), m_fast_sync(true), m_show_time_stats(false), m_sync_counter(0), m_bytes_to_sync(0), m_cancel(false),
   m_long_term_block_weights_window(CRYPTONOTE_LONG_TERM_BLOCK_WEIGHT_WINDOW_SIZE),
   m_long_term_effective_median_block_weight(0),
@@ -872,6 +872,9 @@ uint64_t Blockchain::get_difficulty_for_next_block()
   }
 
   CRITICAL_REGION_LOCAL(m_blockchain_lock);
+  // one snapshot for the tip and every block read below; without it each read
+  // renews its own txn and a batch committing mid-loop splices two chains
+  db_rtxn_guard rtxn_guard(m_db);
   std::vector<uint64_t> timestamps;
   std::vector<difficulty_type_128> difficulties;
   uint64_t height;
@@ -894,7 +897,12 @@ uint64_t Blockchain::get_difficulty_for_next_block()
   //    then when the next block difficulty is queried, push the latest height data and
   //    pop the oldest one from the list. This only requires 1x read per height instead
   //    of doing 735 (DIFFICULTY_BLOCKS_COUNT).
-  if (m_timestamps_and_difficulties_height != 0 && ((height - m_timestamps_and_difficulties_height) == 1) && m_timestamps.size() >= difficulty_blocks_count)
+  // The cached window is keyed by its newest block, not just by height. A
+  // caller that is not the batch writer reads the last committed state, which
+  // during a reorg is the orphaned chain, and would otherwise hand the writer a
+  // window it then extends as if it were current.
+  if (m_timestamps_and_difficulties_height != 0 && ((height - m_timestamps_and_difficulties_height) == 1) && m_timestamps.size() >= difficulty_blocks_count
+      && m_db->get_block_hash_from_height(height - 2) == m_timestamps_and_difficulties_top_hash)
   {
     uint64_t index = height - 1;
     m_timestamps.push_back(m_db->get_block_timestamp(index));
@@ -906,6 +914,7 @@ uint64_t Blockchain::get_difficulty_for_next_block()
       m_difficulties.erase(m_difficulties.begin());
 
     m_timestamps_and_difficulties_height = height;
+    m_timestamps_and_difficulties_top_hash = top_hash;
     timestamps = m_timestamps;
     difficulties = m_difficulties;
   }
@@ -929,6 +938,7 @@ uint64_t Blockchain::get_difficulty_for_next_block()
     }
 
     m_timestamps_and_difficulties_height = height;
+    m_timestamps_and_difficulties_top_hash = top_hash;
     m_timestamps = timestamps;
     m_difficulties = difficulties;
   }
--- a/src/cryptonote_core/blockchain.h
+++ b/src/cryptonote_core/blockchain.h
@@ -1100,6 +1100,10 @@ namespace cryptonote
     std::vector<uint64_t> m_timestamps;
     std::vector<difficulty_type_128> m_difficulties;
     uint64_t m_timestamps_and_difficulties_height;
+    // hash of the newest block in m_timestamps/m_difficulties; the height alone
+    // does not identify the window, since a reorg keeps the height and changes
+    // the contents
+    crypto::hash m_timestamps_and_difficulties_top_hash;
     uint64_t m_long_term_block_weights_window;
     uint64_t m_long_term_effective_median_block_weight;
     mutable uint64_t m_long_term_block_weights_cache_tip_height;


R0BC0D3R:
Verified the cause and the fix against the code.

**Independent confirmation.** I reached the same window position from the arithmetic before seeing this PR. Searching every single-input perturbation of the canonical window that yields 2492059 returns exactly one class: one interior timestamp two seconds later. They are indistinguishable because shifting any interior timestamp by +2s changes two adjacent solvetimes and nets `d sum(solveTime*i) = -2` regardless of position, so `dLWMA = -2/1830`. Height 4377489 sits at index 59 of `[4377430..4377490]`, inside that set. The mechanism described here puts 4377489-A at exactly that index. Same digit, two routes.

**The fix is correct.**

- `get_block_hash_from_height(height - 2)` is the right index: the cached window's tip was at `cached_height - 1 = height - 2`.
- Hash equality at a height implies identical ancestry, so validating the newest entry validates the whole window.
- No underflow. `&&` short-circuits and `m_timestamps.size() >= difficulty_blocks_count` (61) must pass first, which implies `height >= 62`.
- `db_rtxn_guard` is writer-safe. `block_rtxn_start()` returns early for `m_writer` before setting `ret = true`, so `active = false` and the guard never stops the batch txn. Same for a thread that already holds a read txn, so it nests correctly.
- Applied to the incident: writer at 4377491, cached height 4377490, reads 4377489-B from the batch against a stored 4377489-A, mismatch, full rebuild, 2492013.

Worth noting why keying on the hash is stronger than a patch on the RPC path specifically: it validates the cached window against whatever view the caller has, so any stale window fails regardless of how it got published. An RPC thread mid-batch can still cache and revalidate a window off the committed chain, but that only affects the number reported in `get_info` and never reaches disk. The writer reads through the batch txn and rebuilds. The writer is the only one that persists, and the writer is protected.

**Traced the sibling cache; it is clean.** `m_long_term_block_weights_cache_tip_height` has the identical height-only shape (`:1300`, incremental at `:1313`) and `bi_long_term_block_weight` is persisted per block, so it looks like the same exposure. It is not. Every caller is the writer thread or single-threaded startup: `get_next_long_term_block_weight` from `handle_block_to_main_chain` (`:3755`) and `blockchain_import.cpp:453`; `update_next_cumulative_weight_limit` from `init` (`:470`), `pop_block_from_blockchain` (`:668`), `reset_and_set_genesis_block` (`:697`) and `handle_block_to_main_chain` (`:3786`). RPC only reads `get_current_cumulative_block_weight_limit`/`_median`, which return cached members without touching the rolling median. No non-writer populates it, so the race cannot occur there. Recording it because the shape invites the question.

**Two operational notes, neither blocking.**

1. Existing corrupted nodes are not healed by this, but that population is self-limiting and self-identifying. Because `bi_diff_lo` is cumulative, every block after the bad one inherits the offset, so the seed window cannot miss it and the node freezes within ~259 blocks, about 4.3 hours. A node syncing normally today is clean, so there is no quietly-corrupted population to hunt for.

   For a node that does need recovery, the pop has to go below the **first corrupted** height, not the freeze height, which is roughly 259 blocks earlier. In the observed case the freeze was at 4377750 and the first bad record was 4377491: a 500-block pop cleared it, a 300-block pop would not have. Then restart, for the reason in the next note.

2. #133 is complementary rather than redundant. It is why `pop_blocks` alone never recovers a node: after the pop the db is corrected but `m_block_cache` keeps serving the stale `diff_lo`, so the node stalls again at the same height. Observed on mainnet last night, pop alone failed and pop plus restart worked.

ngeojiajun (approved):

PR #138, opened 2026-09-04
Fix/coinbase sum 128

XelisVault:


sn1f3rt:
Superseded by #126, which merged as 1b0d3c4. Thanks for picking this up and working through the review feedback while the original author was unavailable.

The two branches converged on nearly the same change, which is a good sign for both. Diffing this head against master, the only substantive difference left is `wide_emission_amount` / `wide_fee_amount`, and those were dropped on purpose rather than overlooked: nothing in this tree emits a `wide_*` field on any wire. `store_difficulty` never did, and the `wide_difficulty` members in `message_data_structs.h` are internal `difficulty_type_128` working fields that `json_object.cpp` does not serialize. The merged version reports the high half through `emission_amount_top64` / `fee_amount_top64` only, matching the `<name>` + `<name>_top64` shape already used for `cumulative_difficulty`, and generalises `store_difficulty` into `store_128` so both go through one helper.

Worth flagging why this cannot simply be merged on top now: it was branched from 40c2e91, before #128 landed. Its `wallet2.cpp` hunks carry `tx.needed_fee = needed_fee` at both call sites, so merging it would revert the sweep fee fix in 6c2bb30. That is what the conflict is pointing at.

The one thing here that master does not have is your comment on `emission_amount += coinbase_amount - tx_fee_amount`. That constraint is worth recording and the reasoning is preserved in the resolved thread on #126. One correction if it gets added later: `validate_miner_transaction` is stricter than "the coinbase covers its fees" - it requires `base_reward + fee == money_in_use` exactly, and this tree never sets `partial_block_reward`. The exact-equality form is what makes the underflow unreachable.

Separately, the `count > bc_height` bound that neither branch touched is now tracked as #140.

PR #161, opened 2026-09-19
fix(wallet): a stored ring aged out of the sanity check and froze the output

R0BC0D3R:
## Problem

`tx_sanity_check` refuses a ring whose median offset falls below half the current output count. That threshold is relative to the chain as it is now, but the ringdb deliberately returns the same ring every time an output is respent. So a ring that passed when it was built starts failing once the chain grows past it, and retrying cannot help because the retry reuses the same ring. The output becomes permanently unspendable through any daemon doing the check.

## Evidence

Seen on the HF14 testnet over ~38 hours and 20,193 transfers:

| | |
|---|---|
| Rejection rate | 48 of 20,193 (0.24%) |
| Ring median | 11151 on **every** rejection |
| Output count over the same window | 64,307 growing to 100,757 |

The median never moves while the threshold climbs, which is the signature of one stored ring being handed back unchanged. It traced to a single wallet holding one output at index 23775, whose ring was recorded when the chain held about 22,300 outputs.

## Fix

Check the ring in `get_outs` before building on it. If it no longer holds up, discard it with `unset_ring` and pick fresh decoys, up to three attempts, then throw rather than return a transaction the daemon will silently refuse. `tx_sanity_check` is split so the rule can be applied to a candidate ring without a `Blockchain`. This mirrors upstream's `wallet2::get_outs`.

The daemon threshold is untouched. Nerva already sits at 50% where upstream uses 60%, and loosening it further would mask the problem rather than fix it.

## Testing

Verified against the actual stuck wallet, not a synthetic case:

```
median offset index is too low (median is 11151 out of total 105897 offsets)
Ring for this transaction failed the daemon's sanity check, discarding it and picking new decoys
indexes: 23775 45900 69042 72867 ...
built ok: 1523 bytes, fee 0.005389
```

Resubmitting with `do_sanity_checks=true`, the check that had refused this wallet for days, returned `status = OK` and `sanity_check_failed = False`. The 66 XNV it could not move is now spendable, and the build ran through the testnet orchestrator with no regression in ordinary transfers.

## Known limits

* The three-attempt exhaustion path, where every re-roll also fails and the wallet throws `Transaction sanity check failed`, is not covered. It needs a ring that cannot be fixed by re-rolling, which is not reproducible without a patched build.
* The `MWARNING` on discard sits below the wallet's default `*:ERROR`, so users will not see it. Left as is because the event is benign and self-healing, but easy to raise if reviewers prefer it visible.

## Related, not fixed here

The daemon's sanity-check branch in `on_send_raw_tx` returns `Failed` without logging anything, so these rejections leave no trace in the daemon log. That is what made this hard to find. Kept out to keep this change wallet-side, but worth a one-line follow-up.


sn1f3rt (approved):
Checked the threshold against the daemon: `rct_offsets[size - CRYPTONOTE_DEFAULT_TX_SPENDABLE_AGE]` on the cumulative distribution counts exactly the outputs `get_num_mature_outputs(0)` counts, so the wallet and daemon agree on what passes. The empty-distribution, light wallet and unknown key image paths all fall through safely, and the last failed attempt leaves no ring behind.

A few small things:

1. `rct_offsets.clear()` after `unset_ring` is redundant, the inner `get_outs` already clears it on entry.
2. The `cryptonote_core/tx_sanity_check.h` include lands between `common/notify.h` and `common/perf_timer.h`; it belongs after the `common/` group.
3. Pre-existing, but since the line is touched: `<< "offsets)."` is missing a leading space and logs as `105897offsets`. Same for `"indices."` in the unique-indices message.

On the trade-off: rings are not only stored by `set_ring` at build time, `process_outgoing` also records them via `add_rings` when a spend is seen on a chain. If that spend is later reorged out, or the ring came from a forked chain, re-rolling produces a second public ring that shares only the real output with the first, which is exactly the intersection the ringdb exists to prevent. It also discards the rings of every input in the tx, not just the stale one, since the check is aggregate. Upstream accepted the same trade-off and the alternative is a permanently frozen output, so I'm fine with it, but it is worth stating in the PR description so it is a deliberate choice rather than a side effect.

